PR #13, opened 2022-08-24
inform people about the cleaninstall task

NotAShelf:
this should be clear (and concise) enough

### DIFF
--- a/pages/Getting Started/Installation.md
+++ b/pages/Getting Started/Installation.md
@@ -119,8 +119,17 @@ that (`gcc>=12.1.0` or `clang>=15`)
 ```Plain
 git clone --recursive https://github.com/hyprwm/Hyprland
 cd Hyprland
-sudo make install
+sudo make install 
 ```
+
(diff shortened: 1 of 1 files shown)


pages/Getting Started/Installation.md
```diff
@@ -119,9 +119,18 @@ that (`gcc>=12.1.0` or `clang>=15`)
...
-sudo make install
+sudo make install 
 ```
 
+Do note that `sudo make install` will copy the example .desktop file to `/usr/share/wayland-sessions/` directory,
+promptly overriding the existent .desktop file from previous installations.
+
+If you would like to move the file yourself, or if you find yourself running `sudo make install` frequently
+and overriding your previous .desktop file (which is definitely a case for NVIDIA users who run Hyprland
+through a wrapper script); you can use `sudo make cleaninstall`, which will do everything `sudo make install does`
```
vaxerski, line 130: `sudo make install does` lol

NotAShelf: smh

PR #18, opened 2022-09-14
Update _index.md

pechano:
Both Sway and Hyprland can work just fine with VFR, if the monitor is "stepped" into the right refresh rate. I have added instructions on how to do this. It requires a bit of experimenting.

### DIFF
--- a/pages/FAQ/_index.md
+++ b/pages/FAQ/_index.md
@@ -265,6 +265,29 @@ the `.desktop` file to point to your script.
 
 Try disabling VFR with `misc:no_vfr=1`.
 
+If VFR is important to you, you can "step" your monitor into proper VFR high refresh mode with some experimenting:
+
+Turn on VFR with `misc:no_vfr=0`.
+
+Use `hyprctl` multiple times to figure out the needed steps and put them in a script that is executed at startup.
+
+*~/.config/hypr/hyrland.conf*
+
+```plain 
+exec-once=~/.config/hypr/hz.sh #makes vfr work properly
+```
+
+Where hz.sh contains the needed steps e.g.
+```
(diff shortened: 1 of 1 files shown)


vaxerski:
what? explain.

pechano:
On my setup (AOC AG27QXP + Radeon 5700XT) I have noticed that adaptive sync can work just fine under Wayland GNOME (using Mutter), but the two wlroots compositors I have tried had issues out of the box. These two being Sway and Hyprland. When I used Sway I figured out (by accident) that if I switched the refresh rates around a few times, I would get rid of the white flashing normally caused by adaptive sync/vfr options. I tried doing it with the Hyprland equivalent commands and it worked. So I just wanted to share it, in case others have the same issue. I have not seen any mention of a solution anywhere.

vaxerski:
please mention it over [here](https://github.com/hyprwm/Hyprland/issues/351) and let others verify this.

pechano:
I have made a post over on the VFR flash issue. Hopefully this method also works for other people. I know it's very hacky, but it may help someone with more knowledge track down the root cause.

NotAShelf:
@pechano is this still relevant?

pechano:
Yeah I get that, no worries. I think the main issue here is that this bug only happens with certain monitors on certain cards, so it is difficult to reproduce at a meaningful scale. I might be one of the only hyprland users with exactly my setup. It would be great if other users who experience flickering were able to experiment a bit.
But I think this is a bug in that will eventually get fixed through wlroots, and not a specific wm. Until then, I will rely on this small script.

vaxerski:
since the thing above got resolved, I see no point in this anymore.

PR #31, opened 2022-10-17
Correct unit for input:repeat_rate

Smona:
The documentation states it's in milliseconds, but in practice it's currently interpreted as Hz.


vaxerski:
thanks :)

PR #32, opened 2022-10-18
hyprland-wiki: `s/windowrule/windowrulev2`

ghost:
It looks like that sometimes people (even me) gets confused about `windowrulev2` and its example, since at the end of the file it shows that the user should use `windowrule`, when the correct type of example is shown on the top of the *Rules* tab:

https://github.com/hyprwm/hyprland-wiki/blob/a15bae6cb2ca0f61f9abebf48e825c96dd2518fe/pages/Configuring/Window-Rules.md?plain=1#L28-L39
Most people just skip to the bottom for the examples, so let's make their life easier :)


vaxerski:
Add more examples of both v1 and v2 then, with complex examples in v2 including multiple props.

PR #33, opened 2022-10-19
Made the "Manual Build" instructions a little more straight forward.

System-x64:


NotAShelf:
what is the point of adding instructions for building wlroots in your Hyprland directory when Hyprland, if cloned with --recursive, already does it for you on make

PR #40, opened 2022-10-25
show alternative explicit keybinds to switch keyboard layouts through hyprctl

proycon:
This was documented at
https://github.com/hyprwm/Hyprland/wiki/Uncommon-tips-&-tricks, referenced by hyprwm/Hyprland/issues/316, but seemed not to have made it into the new wiki. I think showing these `hyprctl keyword input:kb_layout` commands is still very informative for users, even though the method already documented may be easier and preferred.


vaxerski:
yeah but they are stupid and wrong. You should change per-device keymaps instead.

PR #42, opened 2022-10-30
explained how to use multiple dispatchers with the same keybind

AmmarHaddadi:
yeah just the title, wish you a good day/night


AmmarHaddadi:
dunno what happened but I closed this and reopened multiple times sorry

PR #46, opened 2022-11-04
Added a needed package for obs

AmmarHaddadi:
this package doesn't come as a dependency (technically it's not coz most of ppl use xorg) but even tho OBS  can't work with out 
image


vaxerski:
don't <br/>
just double newline.

AmmarHaddadi:
omg never knew abt this !
image

PR #47, opened 2022-11-08
Add libinput links and reword some settings

Daxtorim:
Clarified the self-explanatory stuff that isn't self-explanatory at all when you don't know about `libinput`. Also moved things around to group settings by keyboard / cursor / scrolling / focus-behavior.

PS: 
Regarding `follow_mouse = 3`, I tried to reword it to what I think the setting should be doing based on the old bullet point, but it doesn't do anything. `3` is just the same behavior as `0`. I don't know if this is a bug or if I just really don't understand what this setting is supposed to do. (I'm still on https://github.com/hyprwm/Hyprland/commit/c9eb0f3aabf4e6ea96aa6a20d2e2b10d19adeb36 as I follow [this COPR package](https://copr.fedorainfracloud.org/coprs/kasion/Hyprland-git/) on fedora, so I don't know if this has changed in the meantime)

### DIFF
--- a/pages/Configuring/Variables.md
+++ b/pages/Configuring/Variables.md
@@ -121,38 +121,44 @@ _[More about Animations](../Animations)._
 
 | name | description | type | default |
 |---|---|---|---|---|
+| kb_model | Appropriate XKB keymap parameter. See the note below. | str | \[EMPTY\] |
 | kb_layout | Appropriate XKB keymap parameter | str | us |
 | kb_variant | Appropriate XKB keymap parameter | str | \[EMPTY\] |
-| kb_model | Appropriate XKB keymap parameter | str | \[EMPTY\] |
 | kb_options | Appropriate XKB keymap parameter | str | \[EMPTY\] |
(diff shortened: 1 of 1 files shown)


vaxerski:
difference between 0 and 3 is that in 0 the mouse cursor will always be passed to the active window, while in 3 it will be passed to whatever window it's on. That means, with 3, you can scroll on a different window (than your kb focused one), while on 0, you can't.

Daxtorim:
Addressed the review comments.

---

> That means, with 3, you can scroll on a different window (than your kb focused one), while on 0, you can't.

Nope, doesn't work here. That is what 2 does, though.

vaxerski:
odd. Means I've prolly broken it

vaxerski:
fixed.

Daxtorim:
> fixed

Can confirm.

Lemmino if you want more changes to this :eyes:

vaxerski:
> Lemmino if you want more changes to this :eyes:

is that a motherfucking lemmino reference

good youtuber

vaxerski:
1. thanks!

PR #55, opened 2022-11-17
Better documentation for fullscreen dispatcher

raffaem:
As currently stated, the doc can be interpreted to mean that 0 makes the window fullscreen and 1 makes it maximized. In reality both of them are toggles.


vaxerski:
not exactly, as using fullscreen on an already maximized/fullscreen'd window will bring it back to normal regardless of the arg.

raffaem:
That's what I wrote, no?

the arg 0 toggle between tiled (what you called "normal") and windowed

the arg 1 toggle between tiled (a.k.a. "normal") and fullscreened

vaxerski:
yeah I just thought the "if already fullscreen 0, what does fullscreen 1 do" was a bit unclear, if you want to be extra clear.

vaxerski:
closing this, not much point

PR #63, opened 2022-11-26
Screenshot utils

raffaem:


fufexan:
How is Hyprshot different from our own Grimblast?

raffaem:
> How is Hyprshot different from our own Grimblast?

Is this question for me?

If so, I don't know. I didn't try Grimblast.

PR #67, opened 2022-11-30
Fix Master-Layout.md

Narcha:
Fixed:
- Invalid markdown syntax on one of the tables
- Default value of `new_is_master` was incorrectly stated as `false`
  See https://github.com/hyprwm/Hyprland/blob/b2cb3b8bf25e42a44dc53681938ec1399f56d784/src/config/defaultConfig.hpp#L97

### DIFF
--- a/pages/Configuring/Master-Layout.md
+++ b/pages/Configuring/Master-Layout.md
@@ -17,9 +17,9 @@ However, you can resize the master window.
 _category name `master`_
 
 | name | description | type | default |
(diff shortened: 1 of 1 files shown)


vaxerski:
wrong line

https://github.com/hyprwm/Hyprland/blob/b2cb3b8bf25e42a44dc53681938ec1399f56d784/src/config/ConfigManager.cpp#L107

but yes

PR #82, opened 2022-12-17
fix layout instead of kb_layout

FlafyDev:
doesn't use the prefix "kb_" for "layout" which results in an error.


vaxerski:
thanks!

PR #92, opened 2022-12-31
Fix typo in App-Clients.md

Genius1512:
Replace "is" by "it's"


vaxerski:
sorry my guy but your grammar sucks. it should be "its"

PR #96, opened 2023-01-06
Update Environment-variables.md

luigir-it:
link broken


vaxerski:
I don't think the .md is correct

NotAShelf:
1. You should not be using file extensions in link references
2. Quick start no longer exists. It should point at the correct subsection in master tutorial

I'll make a PR soon.

PR #102, opened 2023-01-15
add keyword instructions for tablet input devices

ppenguin:
Instructions for new feature from https://github.com/hyprwm/Hyprland/pull/1319 and minor additional clarification/refactor how to use `devices:<name> {}`

### DIFF
--- a/pages/Configuring/Keywords.md
+++ b/pages/Configuring/Keywords.md
@@ -103,15 +103,48 @@ Inside of it, put your config options. All options from the `input` category
 
 force_no_accel, follow_mouse, float_switch_override_focus
 
-Properties that change names:
+Properties that change names according to the device name:
 
 ```plain
-touchdevice:transform -> transform
-touchdevice:output -> output
+<touchdevice>:transform -> transform
+<touchdevice>:output -> output
 ```
+
+where `<touchdevice>` is again the device as obtained by `hyprctl devices`.
+
+Example for tablet with touch screen and stylus:
+```plain
+> hyprctl devices
+...
+Tablets:
+        Tablet at 2a54080:
+                elan2514:00-04f3:29f5-stylus
+
+
+Touch:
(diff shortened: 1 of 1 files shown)


pages/Configuring/Keywords.md
```diff
@@ -103,16 +103,49 @@ Inside of it, put your config options. All options from the `input` category
 
 force_no_accel, follow_mouse, float_switch_override_focus
 
-Properties that change names:
+Properties that change names according to the device name:
 
 ```plain
-touchdevice:transform -> transform
-touchdevice:output -> output
+<touchdevice>:transform -> transform
```
vaxerski, line 109: whats with the `<>`?

I'd just do touchdevice:transform, tablet:transform -> transform

pages/Configuring/Keywords.md
```diff
@@ -103,16 +103,49 @@ Inside of it, put your config options. All options from the `input` category
...
+<touchdevice>:output -> output
 ```
 
+where `<touchdevice>` is again the device as obtained by `hyprctl devices`.
+
+Example for tablet with touch screen and stylus:
+```plain
+> hyprctl devices
```
vaxerski, line 117: usually we do $

vaxerski (changes requested):

pages/Configuring/Keywords.md
ppenguin: the `<>` is to indicate that it's a placeholder for the actual device name. If found that it is sometimes unclear in the wiki where verbatim text is meant and where a placeholder is meant, which is especially confusing for words that could be used as class names. So I believe that we should indicate placeholders as such, if there's a better way *how* I'm all ear ;).

pages/Configuring/Keywords.md
ppenguin: Ok, will change in next commit

PR #117, opened 2023-02-07
update Other.md: Change hyprland-per-window-layout to better fork

coffebar:
Change hyprland-per-window-layout to [another](https://github.com/coffebar/hyprland-per-window-layout) version: with zero configuration, multi-keyboard support and available  to install from package manager


vaxerski:
shameless self-plug

PR #124, opened 2023-02-11
add one more screenshot to examples

coffebar:
mine looks different, hope it will be in this list


vaxerski:
hella ugly but alright some people might like it.

PR #134, opened 2023-02-23
Add FAQ about VRR flickering

jcdickinson:
VRR can cause flickering on some monitors in desktop scenarios. Add a recommendation to change VRR to `2` if this is happening.


vaxerski (approved):
thansk

PR #135, opened 2023-02-23
new solution to disable middle-click paste

BvngeeCord:
`wl-paste -p --watch wl-copy -pc`.
I haven't tested the new solution for *that* long yet, but I'm fairly confident that it works, at least better than the original solution which was already flawed.

### DIFF
--- a/pages/FAQ/_index.md
+++ b/pages/FAQ/_index.md
@@ -266,12 +266,14 @@ to use absolute paths, such as `/home/username/Script` instead of `~/Script`
 
 # How to disable middle-click paste?
 
-In your config, add this bind: `bind = , mouse:274, exec, ;`. Note that the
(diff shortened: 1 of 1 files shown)


BvngeeCord:
> I haven't tested the new solution for that long yet

Coming back to this. It seems to not work in some applications and I have no idea why. Even when I try manually running `wl-copy -pc` to clear the primary buffer. `wl-paste -p` says it's empty, but going back to the application and middle clicking still pastes whatever was selected before. I've noticed this so far ONLY in Discord (canary) and Webcord, which is likely significant. I don't really know where to start or how to debug this. If anyone could help or give suggestions that would be greatly appreciated. Also let me know if I should make a new issue for it.

BvngeeCord:
So far, seems like this only breaks in electron + xwayland apps. Very frustrating, looking into it more

PR #140, opened 2023-02-27
fix link in Using-Plugins.md

coffebar:
I open this link in web and get 404 error, but it can be opened without ".md" just fine


vaxerski:
it's already fixed

PR #145, opened 2023-03-04
Fix for eww listeners not updating properly

matthis-k:
I've noticed the listeners do not update while their values are only present in the class attribute, so either the scripts would have to be merged to one output or one can use a hidden element like this to keep the variables updated. For further context see [this issue on eww](https://github.com/elkowar/eww/issues/669)


vaxerski:
@fufexan

fufexan:
I would be in favour of having a single script instead of the current 3, but this will do.

PR #149, opened 2023-03-07
update plugins guide

Selmer443:
added some more examples with details, moved the disclaimer to the top of the page

### DIFF
--- a/pages/Plugins/Using-Plugins.md
+++ b/pages/Plugins/Using-Plugins.md
@@ -2,37 +2,7 @@ This page will tell you how to use plugins.
 
 {{< toc >}}
 
-## Getting plugins
-
-Plugins come as _shared objects_, aka. `.so` files.
-
-Hyprland does not have any "default" plugins, so any plugin you may want
-to use you will have to find yourself.
-
-## Installing / Using plugins
-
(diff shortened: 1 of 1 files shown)


pages/Plugins/Using-Plugins.md
```diff
@@ -22,7 +33,35 @@ In rare cases, they might even need to be compiled on the same machine.
...
+cd Hyprland && make protocols
+```
+
+You will also need to set the HYPRLAND_HEADERS env variable to the path 
+of the cloned Hyprland repository (not the Hyprland-Plugins directory).
+You are now ready to compile a plugin! 
+
+Example compilation of borders-plus-plus Hyprland plugin:
+```sh
+cd hyprland-plugins/borders-plus-plus && make
```
vaxerski, line 63: why not set the envvar here as an example?

vaxerski (changes requested):
A note about recompiling plugins after a Hyprland update would also be nice if we're at that. Also mention to update the cloned source.

Furthermore, we need a note for checking out hyprland if someone's using a release and not -git.

vaxerski:
with 30d164e00dec486cdbefaa0af0ef542c3f04b041 I think this is not relevant anymore

PR #150, opened 2023-03-08
Update mouse cursor documentation

milkknife:
I was having trouble with getting the cursor to work on chromium/electron flatpak applications. I updated this section of the documentation with what worked for me.


vaxerski (changes requested):
I'd leave the arch wiki link part JIC at the bottom, but yes it shouldnt be necessary.

vaxerski (approved):
lgtm thanks!

PR #153, opened 2023-03-09
Missing 'exec' keyword added to the clipman manual.

szostah:
I've noticed that a dispatcher keyword is missing in the clipman manual.


vaxerski (approved):
thanks! :)

PR #155, opened 2023-03-13
Add animations/blur/etc toggle script

milkknife:
I was getting worse performance than sway with wine applications, and I would like to share what fixed this for me.


vaxerski (approved):
thanks!

PR #156, opened 2023-03-13
Update switchxkblayout section

milkknife:
I spent a bit longer than I should have figuring out that for hyprland, just the device name alone is enough, whereas on sway you would need `input:keyboard` or `input:DEVICE_NAME`. Also wanted to share a tip for anyone who uses several layouts with different variants or global options on each.


vaxerski (approved):
lgtm thanks

PR #162, opened 2023-03-20
Change where to find plugins link to point to the github repo than duckduckgo

devadathanmb:
As mentioned in the title, I've changed the link in the [Where to find plugins](http://wiki.hyprland.org/Plugins/Using-Plugins/#where-do-i-find-plugins) section to point to the plugins repo rather than duckduckgo.


vaxerski:
why though? Only _official_ plugins are there, and e.g. hyprload or split-workspaces arent.

vaxerski:
397d73723d4696c73744323e4c9d4cfeffd6d840

PR #166, opened 2023-03-24
Add volume limit to keybinds

FoxWDN:
wpctl seems to increase the volume without any limit by default, which *might* damage  earphones or make ears hurty, -l 1.5 option limits this to sane levels.


vaxerski:
thanks!

PR #172, opened 2023-04-01
Did some minor changes to the wiki

r454354:
Did some minor changes to the wiki - 
1) Added wikipedia link to the window rule regex topic so that people who are not well versed with programming can understand what are regexes. Wikipedia explained regex in a simple and lucid english so that anyone can understand.
2) Many people are facing graphical glitches in games and other applications which launches full screen by default. Setting no_direct_scanout to true really works. See this discord message link where I helped the guy who was having the same problem - https://discord.com/channels/961691461554950145/961694014615208048/1091369483811160125 . See this video - https://0x0.st/HHLn.mp4
3) Added the regex cheatsheet link. Awesome github repo it explains regex in details ( basic , characters , meta characters , Shorthand Character Sets , Lookarounds , Flags , Greedy vs Lazy Matching etc etc ).

### DIFF
--- a/pages/Configuring/Variables.md
+++ b/pages/Configuring/Variables.md
@@ -241,7 +241,7 @@ Described [here](../Keywords#per-device-input-configs).
 | animate_mouse_windowdragging | If true, will animate windows being dragged by mouse, note that this can cause weird behavior on some curves | bool | false |
 | disable_autoreload | If true, the config will not reload automatically on save, and instead needs to be reloaded with `hyprctl reload`. Might save on battery. | bool | false |
 | enable_swallow | Enable window swallowing | bool | false |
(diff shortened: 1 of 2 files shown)


vaxerski:
On this note, we could also add an FAQ entry about the graphical glitches.

r454354:
> On this note, we could also add an FAQ entry about the graphical glitches. Tho TBF, I am thinking about defaulting no_direct_scanout to true

Imo defaulting no_direct_scanout to true would be a nice option. Those who don't want this feature can disable. It is better not to add extra FAQ entry about the graphical glitches. Hyprland newbies never read the faq section of the wiki.

vaxerski:
Fair. Thansk!

vaxerski (approved):
lgtm

PR #173, opened 2023-04-04
chore(docs): update hidpi config changes

ndom91:
Was having issues getting the hidpi settings to really have any effect. Even after installing the aur hyprland-hidpi package with all the required patches and git reverts, modifying my `.Xresources` `dpi` setting, etc.

What finally helped was this random comment on a [reddit thread](https://www.reddit.com/r/hyprland/comments/10uzsaf/issue_with_resolution/) about modifying the `monitor` setting from `preferred` to `highres`.

Would be helpful to have in the wiki to save other folks time searching imo.

### DIFF
--- a/pages/Configuring/XWayland.md
+++ b/pages/Configuring/XWayland.md
@@ -31,6 +31,10 @@ and [Pacman patching](https://wiki.archlinux.org/title/Patching_packages).
 3. Add these lines to your configuration:
 
     ```ini
+
+    # change monitor to hires, the last argument is the scale factor
+    monitor=,highres,auto,2
+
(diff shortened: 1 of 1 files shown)


vaxerski:
@fufexan you wrote all that

fufexan:
Right. It seemed obvious to me that you would only try using HiDPI XWayland if you already had Hyprland running scaled. Sure, let's get it in.

ndom91:
Yeah, while you're right for someone experienced with Hyprland or maybe wayland more generally. But I had just switched to both a few days ago and didn't recognize this was an option. 

Thanks for the quick turnaround!

PR #174, opened 2023-04-05
Add wl-screenrec as suggestion for screen recording

russelltg:
Disclosure: I am the author of wl-screenrec

Add a link to wl-screenrec, a more efficient version of wf-recorder that works for Intel and AMD GPUs.


vaxerski:
thanks!

PR #178, opened 2023-04-09
Added some minor stuff again

MrDisguised:
Again added some minor stuff to the wiki - 

1) It was important to add the [`wl-clip-persists`](https://github.com/hyprwm/hyprland-wiki/commit/765e1a4ab2499ac9554e1234a35758b397440411) program to the wiki. Normally we use wl-clipboard in wayland to store the copied data (for ex text,images) to our clipboard but there is a problem with it. When we close the application from where we copied the data from the data disappears from the clipboard. For example lets say u are using firefox and discord you copy a text from firefox and paste it in discord. Everything works well !! but now try to copy a text from firefox then close the firefox application and now try to paste the text you copied from firefox into discord (or any other application). It doesn't work :cry: . So `wl-clip-persist` is used so that it can store the data in our clipboard even after the applications exits out.

EDIT: I know that [wl-clip-persist](https://github.com/Linus789/wl-clip-persist) is a niche project but it works!! I couldn't find a better alternative for it.

2) [Special workspace](https://wiki.hyprland.org/Configuring/Dispatchers/#special-workspace) section was not explained properly. Users should know how a special workspace works thats why I gave an [example](https://github.com/hyprwm/hyprland-wiki/commit/709e189094a314629ef63563472e3bf7c8291264) so that people can understand it better.

### DIFF
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -128,6 +128,14 @@ will send it to the currently active _real_ workspace.
 You can define multiple named special workspaces, but the amount of those is limited to 97 at a time.
 {{< /hint >}}
 
+For example, to move a window/application to a special workspace you can use the following syntax:
+
+```
+bind = SUPER, C, movetoworkspace, special
+#The above syntax will move the window to a special workspace upon pressing 'SUPER'+'C'.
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
lgtm, thanks!

PR #180, opened 2023-04-10
Update Must-have.md

SingleSlice:
While using opensuse, i found out ( im a bit of a newbie ) that libexec is in a different path, so i thought of adding it in the docs.


vaxerski (approved):
lgtm

PR #181, opened 2023-04-10
Add socat `-U` flag to IPC example

beni69:
As a temporary fix to https://github.com/hyprwm/Hyprland/issues/1564 I think it would be great to add this to the wiki.

I've just added the flag, but I could also write some explanation for it if necessary.


vaxerski:
fair nuff

PR #187, opened 2023-04-13
clear-up nixos installation

viperML:
I find the NixOS page quite confusing the first time you see it.

And now that the module has been upstreamed into nixpkgs, it could be reasonable to simplify the wiki page.

This is a draft, I want to clean-up the other sections too.

@fufexan


MintzyG:
I agree, the install page is very confusing, I'm lost trying to install it here :(

viperML:
One of my main concerns is regarding the HM module, so I will dump some of my thoughts:

Should we encourage everybody to at least enable the NixOS module before enabling the HM module?
If so, the installation is awkward, because you need the upstream flake just for the HM module.
Also, the NixOS module would provide a different package (HL from nixpkgs) than the HM module (HL upstream)
If not, is the HM module enough to properly run HM? What extra steps would be needed?

Should we move the HM module to HM ([home-manager#3837](https://github.com/nix-community/home-manager/pull/3837))? If so, the installation could be much simpler, as users are not required to pull the upstream flake, and the HM and NixOS module would (potentially) use the same HL version to avoid conflicts. Basically, the same installation path than for Sway.

How should we provide a deprecation of the upstream {NixOS,HM} module then?

fufexan:
> Should we move the HM module to HM (https://github.com/nix-community/home-manager/pull/3837)?

That would be the idea. I've pinged Mic92 as he said in spikespaz's HM rework PR, but I got no response yet from anyone but ncfavier who isn't sure we should upstream (hence my comment about HM moving slow).

In the meantime I'd say we continue having users use the flake for the HM module. We can make the default package be `pkgs.hyprland` instead.

wuyoli:
Are there any plans of making the [hyprland plugins](https://github.com/hyprwm/hyprland-wiki/pull/187) via nix (I see there is already a flake.nix in there, but maybe there could be a section in the docs on installing plugins via Nix?)

fufexan:
@wuyoli indeed, it should be added to the wiki. It was added here https://github.com/hyprwm/Hyprland/pull/2211, but we haven't written docs for it (silly us).
@outfoxxed do you think you'd have some time to make a PR with some docs?

outfoxxed:
@fufexan yeah I forgot about that. Will do.

fufexan:
@viperML do you want to continue working on this?

viperML:
No

(I'm on sway..)

fufexan:
Okay, I'll take it up from here then.

pages/Getting Started/Installation.md
```diff
@@ -38,7 +38,19 @@ hyprland - binary x86 tagged release
...
+
+{{< tab "Nix" >}}
+
+Enable Hyprland in your NixOS configuration:
+
+```nix
+programs.hyprland.enable = true;
+```
+
+For more details, read the [Nix page](../../Nix).
```
NotAShelf, line 50: Lets maybe mention that Hyprland is now in nixpkgs, and this option only enables the module from nixpkgs and not the flake, which would render a Hyprland input semi-useless. People have a tendency to copy others configurations and confuse themselves.

pages/Nix/Hyprland on Home Manager.md
```diff
@@ -62,3 +71,34 @@ in {
...
+## Plugins
+
+Hyprland plugins can be added through an option:
+
+```nix
+wayland.windowManager.hyprland.plugins = [
+  inputs.hyprland-plugins.packages.${pkgs.system}.hyprbars
+  "/absolute/path/to/plugin.so"
+];
+```
```
NotAShelf, line 101: What about plugins not providing a flake? I think we can give an example of a builder here, which generates a plugin as the output.

fufexan: That'll be in a separate PR. I also want to make a general builder and a flake template, and then I'll also add instructions here.

pages/Getting Started/Installation.md
fufexan: That's what the "more details" is for.

PR #192, opened 2023-04-15
Update FreeBSD dependencies

jbeich:


vaxerski:
how is jq a hard dep?

jbeich:
```
$ meson setup hyprland_build hyprland
The Meson build system
Version: 1.1.0
Source dir: /tmp/hyprland
Build dir: /tmp/hyprland_build
Build type: native build

hyprland/meson.build:2:12: ERROR: Program or command 'jq' not found or not executable

A full log can be found at /tmp/hyprland_build/meson-logs/meson-log.txt
```

vaxerski:
huh. Meson. Maybe @fufexan did something. Fair.

vaxerski (approved):
thanks!

fufexan:
> huh. Meson. Maybe @fufexan did something. Fair.

Yeah, remember when we implemented version reading from `props.json`?

vaxerski:
right.

eli-schwartz:
> Yeah, remember when we implemented version reading from `props.json`?

Sounds incredibly painful, why make it into json instead of just a version file containing a single line with the version?

fufexan:
> > Yeah, remember when we implemented version reading from `props.json`?
> 
> Sounds incredibly painful, why make it into json instead of just a version file containing a single line with the version?

I don't remember the real reason, but I think we wanted to be future proof in case another variable needed to be added.

eli-schwartz:
Has this turned out to be a problem in practice? What other information might need to be read into the build system itself?

fufexan:
> Has this turned out to be a problem in practice? What other information might need to be read into the build system itself?

No. We just over engineered it seems. 🤷‍♂️

PR #194, opened 2023-04-22
Fix shellcheck warnings and optimize code blocks

Ordoviz:
I noticed that the wiki sometimes uses Bashism in `#!/bin/sh` blocks, so I ran [markdown-shellcheck](https://github.com/Ordoviz/markdown-shellcheck) on it to find more issues.

### DIFF
--- a/pages/Configuring/Expanding-functionality.md
+++ b/pages/Configuring/Expanding-functionality.md
@@ -12,8 +12,8 @@ to react to different events. See its description
 This bash script will change the outer gaps to 20 if the currently focused
 monitor is DP-1, and 30 otherwise.
 
-```sh
-#!/bin/sh
+```bash
+#!/bin/bash
 
 function handle {
   if [[ ${1:0:10} == "focusedmon" ]]; then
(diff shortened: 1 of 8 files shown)


pages/Configuring/Expanding-functionality.md
```diff
@@ -12,8 +12,8 @@ to react to different events. See its description
 This bash script will change the outer gaps to 20 if the currently focused
 monitor is DP-1, and 30 otherwise.
 
-```sh
-#!/bin/sh
+```bash
+#!/bin/bash
```
vaxerski, line 16: why are we doing bash here? some people don't use bash

vaxerski (changes requested):

pages/Configuring/Expanding-functionality.md
Ordoviz: Substring expansion `${1:0:10}` is a Bash-only feature

vaxerski: zsh handles it well too

Ordoviz: I have /bin/sh symlinked to /bin/dash so I won't be able to run the script if it uses non-POSIX features despite advertising itself as POSIX shell script with `#!/bin/sh`

vaxerski: technically this is only an example and not a production script, eh, whatever

fufexan (approved):
LGTM

vaxerski (approved):
ok

PR #195, opened 2023-04-22
Easier nvidia setup

kaii-lb:
just makes it a little less annoying to get up and running with hyprland and nvidia.  covers systemd-boot and grub since theyre the most used (i think).


vaxerski (approved):
lgtm thanks

PR #200, opened 2023-04-25
Installation: improve FreeBSD instructions for Manual Build

jbeich:


jbeich:
Some meta notes:
- I don't like CMake, so don't expect me to fix Hyprland (unlike other hyprwm projects)
- Clang/libc++ 16 in FreeBSD **base** (unlike ports) is [currently WIP](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271047) and may take ~1 year to propagate into any -RELEASE.

vaxerski:
huh why is cmake linux only?

jbeich:
> huh why is cmake linux only?

See https://github.com/hyprwm/Hyprland/pull/2209

fufexan (approved):

fufexan:
@jbeich has anything changed since 4 months ago?

jbeich:
Not much:
- Both CMake and Meson instructions still work
- FreeBSD 13.3/14.0 haven't been released yet, so building with GCC is still easier

105b55239eba added `meson subprojects update` but it isn't very useful compared to explicit `git clone --recursive` (or `git submodule update --init --recursive`):
```
$ meson subprojects update --reset
Updating udis86...
  -> Cannot update subproject with no wrap file
Updating tracy...
  -> Cannot update subproject with no wrap file
Updating hyprland-protocols...
  -> Cannot update subproject with no wrap file
Initialized empty Git repository in /tmp/Hyprland/subprojects/wlroots/.git/
remote: Enumerating objects: 480, done.
remote: Counting objects: 100% (480/480), done.
remote: Compressing objects: 100% (466/466), done.
remote: Total 480 (delta 27), reused 160 (delta 5), pack-reused 0
Receiving objects: 100% (480/480), 773.01 KiB | 300.00 KiB/s, done.
Resolving deltas: 100% (27/27), done.
From https://gitlab.freedesktop.org/wlroots/wlroots
 * branch            c74f89d4f84bfed0284d3908aee5d207698c70c5 -> FETCH_HEAD
HEAD is now at c74f89d Avoid using memcpy() to copy structs
Applying diff file "wlroots-meson-build.patch"
Updating wlroots...
  -> c74f89d (grafted, HEAD) Avoid using memcpy() to copy structs [Simon Ser]

$ meson setup build
$ meson compile -C build
[...]
../subprojects/wlroots/include/wlr/render/egl.h:19:10: fatal error: 'wlr/config.h' file not found
#include <wlr/config.h>
         ^~~~~~~~~~~~~~

$ find . -name config.h
./build/subprojects/wlroots/include/wlroots/config.h
./build/subprojects/wlroots/include/config.h

$ cat subprojects/packagefiles/wlroots-meson-build.patch
[...]
-install_subdir('wlr',
+install_subdir('wlroots',
[...]
```

jbeich:
> Both CMake and Meson instructions still work

From this PR, otherwise Meson fails.

fufexan:
> but it's broken outside of Nix

Can I fix it somehow or is it better to just not use `meson subprojects`?

jbeich:
Better remove `wlroots.wrap`. It's not consistent with other subprojects (e.g., hyprland-protocols, udis86) or CMake support.

Downstream I use `git submodule` instead of `meson subprojects` and system wlroots (a la Nix), so don't need synthetic `soversion` bump.

fufexan:
Alright, I'll remove it soon (unless I forget).

jbeich:
Rebased and adjusted dependencies to the current Hyprland and FreeBSD versions.

fufexan:
Is there anything left to do in this PR or is it ready? @jbeich

jbeich:
Tested both Meson and CMake instructions - still good.

jbeich:
Wait, `--skip-subprojects` may be wrong for the plugin API.

fufexan:
Is there anything that can be done on the meson side, so we don't have to include all the commands here? The wlroots.wrap file used to patch the subproject so it properly installed headers, but now I don't know of a better approach.

jbeich:
I've dropped Meson workaround and rebased after aquamarine switch.

pages/Getting Started/Installation.md
```diff
@@ -370,10 +370,11 @@ installed._
...
 
 ```plain
-meson subprojects update --reset
+git clone --recursive https://github.com/hyprwm/Hyprland
+cd Hyprland
 meson setup build
-ninja -C build
-ninja -C build install --tags runtime,man
+meson compile -C build
+meson install -C build --skip-subprojects
```
fufexan, line 377: Does this not mean udis86 will not be found?

jbeich: udis86 is not part of Hyprland plugin API unlike wlroots was before aquamarine. `--skip-subprojects` is still necessary to avoid installing hyprland-protocols (plugins use preprocessed headers instead). Hyprland itself doesn't support building against system-wide copy of udis86 ([Alpine patches this](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/hyprland/use-system-installed-udis86.patch)), and the bundled version uses CMake to build the submodule.

fufexan: That's fine by me. However, I've made progress towards allowing Hyprland to build with a system-wide copy of udis86 (provided it's canihavesomecoffee's fork, which includes a .pc file). See this PR for more info about it, maybe leave your thoughts as well https://github.com/hyprwm/Hyprland/pull/7885.

PR #202, opened 2023-04-27
Add tip about hprctl activewindow

Schievel1:
People have some trouble figuring this out it seems. Since the known tools from X are not working anymore.


vaxerski:
why not just recommend hyprctl clients?

Schievel1:
> why not just recommend hyprctl clients?

You're right, I will reword this a bit to mention that as well. If you have many windows open `hyprctl clients` can become a bit convoluted.

vaxerski (changes requested):

PR #204, opened 2023-04-29
Add tip to install `linux-headers` before running `mkinitcpio`

loqusion:
I kept getting the error `module not found: '***'` for everything I added to my `MODULES` until I used [this solution](https://bbs.archlinux.org/viewtopic.php?id=277580).


vaxerski (approved):
lgtm

PR #206, opened 2023-04-29
Add "How to launch" section under each status bar tutorial

ardishco-the-great:
I'm not sure how to use Eww, so It would be better if someone with better understanding of that added a "How to launch" section under the Eww section.


vaxerski (approved):
fair

PR #215, opened 2023-05-04
Removed the need for XDPH to pull latest git for gentoo

WasteOfO2:
XDPH doesn't need ** keyword, it means portage will pull in the latest git version of XDPH available, which at the time of writing the wiki may have been required, but is not necessary now.

Removing ** will just pull the latest available package available(i.e assume ~amd64) on the gentoo overlay, which is sufficient.


vaxerski (approved):
gangsta

PR #221, opened 2023-05-10
Document home manager plugins array

outfoxxed:
Companion pr for hyprwm/Hyprland#2211

### DIFF
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -97,6 +97,8 @@ You have eight choices:
 
 - Relative workspace on monitor: e.g. `m+1`, `m-1` or `m+3`
 
+- Relative workspace on monitor including empty workspaces: e.g. `r+1` or `r-3`
+
 - Relative open workspace: e.g. `e+1` or `e-10`
 
 - Name: e.g. `name:Web`, `name:Anime` or `name:Better anime`
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -97,8 +97,6 @@ You have eight choices:
 
 - Relative workspace on monitor: e.g. `m+1`, `m-1` or `m+3`
 
-- Relative workspace on monitor including empty workspaces: e.g. `r+1` or `r-3`
-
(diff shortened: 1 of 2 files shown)


viperML:
This doesn't explain how to get the plugin packages

pages/Configuring/Dispatchers.md
```diff
@@ -97,8 +97,6 @@ You have eight choices:
 
 - Relative workspace on monitor: e.g. `m+1`, `m-1` or `m+3`
 
-- Relative workspace on monitor including empty workspaces: e.g. `r+1` or `r-3`
```
fufexan, line 100: Why remove this? I don't see how it's related to this PR.

pages/Nix/Options & Overrides.md
```diff
@@ -29,6 +29,17 @@ programs.hyprland = { # or wayland.windowManager.hyprland
 };
 ```
 
+### HM module
```
fufexan, line 32: This should go below `## Options descriptions`, before or after `### XWayland`. It should be something like

    ### Plugins

    ```nix
    wayland.windowManager.hyprland = {
      plugins = [
        inputs.hyprland-plugins.packages.${pkgs.system}.hyprbars # package
        "/home/<user>/Projects/my-hyprland-plugin/libmy-hyprland-plugin.so" # path
      ];
    };
    ```

And then you should include more info about how to build plugins, like "look at how pacakges are defined in `hyprland-packages`".

fufexan (changes requested):

pages/Configuring/Dispatchers.md
outfoxxed: No clue how that got in here. Good catch.

pages/Nix/Options & Overrides.md
outfoxxed: Alright

outfoxxed: Done. You probably want to squash this.

fufexan:
Okay, it looks good enough for now. The Nix part of the wiki will need a big rewrite anyway once I can get a plugin builder in Nix.

viperML (approved):

PR #222, opened 2023-05-17
Improve eww get-active-workspace for multimonitor setup

Fymyte:


vaxerski (approved):
lgtm, but I'll ping @fufexan for a check too

fufexan:
Are you sure jq doesn't need `-r` to print without quotes? Or does it not affect the script?

Fymyte:
> Are you sure jq doesn't need `-r` to print without quotes? Or does it not affect the script?

Since the id is a number there is no quotes printed in the output. I can re-check in 2h if needed.

fufexan:
It's alright, I'll merge it as is.

PR #224, opened 2023-05-21
add xwayland screensharing documentation

Rdeisenroth:
Recently learned about [xwaylandvideobridge](https://invent.kde.org/system/xwaylandvideobridge) which is a very user friendly way to enable screensharing of Wayland applications and Screens for Apps runnuing under XWayland. 

Since it takes some workarounds to get it working with Hyprland i thought it would be nice to have it documented here since it's currently not described in the gist. 

Please feel free to correct me if i put in some misinformation about Hyprlands ability to hide windows but this was the easiest way i could find.


vaxerski (approved):
gangsta

PR #229, opened 2023-05-26
Replace wrapper mention from faq page

gurka420:
Wrappers have been deprecated since 0.23.0, but were not removed from the faq page.
I am not very efficient in markdown, feel free to correct the formatting if there are errors.


vaxerski (approved):
right, thanks!

PR #231, opened 2023-05-27
fix typo in screen sharing page

zaxutic:
I assume we're not trying to archive the effect


vaxerski (approved):
archiving effects since 1984

PR #232, opened 2023-05-27
Add more details about combining extra args for monitor rule

Fymyte:
Explicitly say that you can combine extra args for monitor rules


vaxerski (approved):
sure

PR #241, opened 2023-06-04
Fix: Improve XDP(s) processes termination.

juangiordana:
Improve `xdg-desktop-portals` processes termination since `xdg-desktop-portals` is 19 characters long.

From `killall(1)` man page:

> -e, --exact
>
> Require an exact match for very long names. If a command name is longer
> than 15 characters, the full name may be unavailable (i.e. it is swapped
> out). In this case, `killall` will kill everything that matches within
> the first 15 characters. With -e, such entries are skipped. `killall`
> prints a message for each skipped entry if -v is specified in addition
> to -e.

### DIFF
--- a/pages/FAQ/_index.md
+++ b/pages/FAQ/_index.md
@@ -235,7 +235,7 @@ In such cases, a script like this:
 ```bash
 #!/bin/bash
 sleep 4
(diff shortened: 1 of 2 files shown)


vaxerski (approved):
ok

PR #244, opened 2023-06-05
Fixed wiki having wrong type and default for groupbar_text_color

BrianCArnold:


vaxerski (approved):
lol my bad, epic c+p fail

PR #254, opened 2023-06-22
Add hyprland instance signature to handle_monitor_connect.sh

skogler:
When I tried this tip from the wiki to handle monitor connects, I discovered that the event socket is now namespaced with the instance id.

This makes this tip work with version 0.26.0-3 (arch package).


vaxerski (approved):
oh god yeah that was ooooold

PR #256, opened 2023-06-23
updating window rule regex documentation for more clear help

thatONEjustin:


pages/Configuring/Window-Rules.md
```diff
@@ -35,13 +35,25 @@ the `RULE` field is unchanged, but in the `WINDOW` field, you can put regexes
...
+
+for the `windowrulev2 = float,class:(kitty),title:(kitty)` example, the `class:(kitty)` `WINDOW` field is what keeps the window rule
+specific to kitty terminals. 
+{{< /hint >}}
+
 For now, the supported fields are:
 
 ```ini
-class - class regex
+class - class regex
```
vaxerski, line 56: added by mistake?

pages/Configuring/Window-Rules.md
```diff
@@ -35,13 +35,25 @@ the `RULE` field is unchanged, but in the `WINDOW` field, you can put regexes
 for multiple values like so:
 
 ```ini
-windowrulev2 = float,class:^(kitty)$,title:^(kitty)$
+windowrulev2 = float,class:(kitty),title:(kitty)
```
vaxerski, line 38: why remove the ^ and $?

pages/Configuring/Window-Rules.md
thatONEjustin: yeah this one is by mistake

pages/Configuring/Window-Rules.md
thatONEjustin: the example kind of only specifically works for a kitty terminal window that doesn't dynamically update the title. This is because ^ tells regex to look for the needle "at the beginning of the string". $ tells regex to start at the eend. 

If you used this same regex with Alacritty with dynamic titles the window rule would never apply.

example:

`windowrulev2 = float, class:^(alacritty),title:^(alacritty)$` 

Will never match even a window title of: `alacritty - neofetch` because the search needle isn't present at both the beginning and end of the string.

vaxerski: float is not a dynamic windowrule and kitty initially always has "kitty", no?

thatONEjustin: not the actual float rule, I mean the title of a window.

the regex in the wiki ONLY works if the window's title is "kitty" and nothing else. If the terminal window for whatever reason were to change it's title dynamically, the rule would no longer apply. 

Kitty itself isn't the best example for what I'm trying to clarify, as you're saying it's class/title is always initially "kitty".

Looking at the titles of a Mozilla Firefox browser is easier because each tab has it's own title that would have to match the rule exactly.

vaxerski: meh, I just found ^$ as less confusing, people now will wonder wtf is ()

PR #260, opened 2023-06-27
Add Void Linux Installation tab

Makrennel:
It isn't available in the official Void Linux repositories, but I've created template files which can build it using xbps-src consistently: https://github.com/Makrennel/hyprland-void

I figured it'd be good to share it given how many useless issues have been made regarding void linux build problems.


vaxerski (approved):
void more like void of any redeeming qualities amirite :trollface:

PR #262, opened 2023-07-02
Add pure evaluation (stub) to `Hyprland on Home Manager.md`

harukaWinter:
As title

### DIFF
--- a/pages/Nix/Hyprland on Home Manager.md
+++ b/pages/Nix/Hyprland on Home Manager.md
@@ -66,22 +66,16 @@ in {
 ```nix
 { pkgs, ... }:
 let
-  flake-compat = builtins.fetchGit {
-    name = "flake-compat";
-    url = "https://github.com/edolstra/flake-compat";
-    ref = "refs/heads/master";
-    rev = "35bb57c0c8d8b62bbfd284272c928ceb64ddbde9";
+  flake-compat = builtins.fetchTarball {
+    url = "https://github.com/edolstra/flake-compat/archive/35bb57c0c8d8b62bbfd284272c928ceb64ddbde9.tar.gz";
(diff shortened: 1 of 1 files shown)


fufexan:
It's a lot easier to just modify the URL to a commit and add the extra `sha256` to `fetchTarball` instead of using `builtins.fetchGit`. `builtins.fetchGit` is usually not used due to it downloading the entire repo, whereas something like `pkgs.fetchFromGitHub` uses `builtins.fetchTarball` under the hood to minimize the amount of data downloaded. Also, the user should only care about files' states at one particular commit, not the entire history.

pages/Nix/Hyprland on Home Manager.md
```diff
@@ -62,3 +62,37 @@ in {
   };
 }
 ```
+### Pure evaluation
+```nix
+{ pkgs, ... }:
+let
+  flake-compat = builtins.fetchGit {
```
fufexan, line 69: Should be
```nix
  flake-compat = builtins.fetchTarball {
    url = "https://github.com/edolstra/flake-compat/archive/master.tar.gz"; # optionally replace with commit
    sha256 = ""; # see comment below
  };

  hyprland = (import flake-compat {
    src = builtins.fetchTarball {
      url = "https://github.com/hyprwm/Hyprland/archive/master.tar.gz"; # optionally replace "master" with a commit
      sha256 = ""; # not sure the builtin works like this, may need `lib.fakeHash256` instead of the empty string
    };
  }).defaultNix;
  ```

harukaWinter:
@fufexan Sorry for the late reply. I've [replaced fetchGit with fetchTarball](https://github.com/hyprwm/hyprland-wiki/pull/262/commits/a6c6672e4a6de548d1f5ed2a35652c2ee25334b9) and it works on my machine!

fufexan:
We may not want to hardcode commits and hashes, but rather instruct the user to add them by themselves.

PR #266, opened 2023-07-10
dwindle: document smart_split

DanielAdolfsson:
This adds some docs for the smart_split option introduced in hyprwm/Hyprland#2676


vaxerski (commented):
quadrant is the default. Smart split uses triangles.

DanielAdolfsson:
Think this should make it bit clearer :)

vaxerski (approved):
better

PR #273, opened 2023-07-21
Nix/HM: add instructions for upstream module

### LINKED ISSUE #272: add official home-manager module to docs
as of https://github.com/nix-community/home-manager/commit/ee5673246de0254186e469935909e821b8f4ec15, there is an official home-manager module for hyprland. having both the official and unofficial modules at the same time can cause issues

fufexan:
Adds instructions on how to use the upstream Home Manager module for Hyprland.
Will probably undergo some changes as the HM module needs a few changes.

Fixes #272

### DIFF
--- a/pages/Nix/Hyprland on Home Manager.md
+++ b/pages/Nix/Hyprland on Home Manager.md
@@ -1,7 +1,5 @@
-You can use the Home Manager module by adding it to your configuration:
-
 For a list of available options, check the
-[module file](https://github.com/hyprwm/Hyprland/blob/main/nix/hm-module.nix).
+[Home Manager options](https://nix-community.github.io/home-manager/options.html#opt-wayland.windowManager.hyprland.enable).
 
 {{< hint title=Note >}}
 - *(Required) NixOS Module*: enables critical components needed to run Hyprland properly
@@ -10,13 +8,24 @@ For a list of available options, check the
 
 ## Installation
 
-The following snippets of code try to show how to bring the Hyprland flake from
-the flake input and import it into the module system. Feel free to make any
-adjustment for your setup.
-
 {{< tabs "uniqueid" >}}
 
+{{< tab "Home Manager" >}}
+
+Home Manager has options for Hyprland without needing to import the Flake module.
+
+```nix
+{
+  wayland.windowManager.hyprland.enable = true;
+}
+```
+
+{{< /tab >}}
 {{< tab "Flakes" >}}
+
+The following snippet of code tries to show how to bring the Hyprland flake from
+the flake input and import it into the module system. Feel free to make any
+adjustment for your setup.
 
 Don't forget to replace `user@hostname` with your username and hostname!
 
@@ -51,6 +60,12 @@ Don't forget to replace `user@hostname` with your username and hostname!
 {{< /tab >}}
 
 {{< tab "No flakes (with flake-compat)" >}}
+
+The following snippet of code tries to show how to bring the Hyprland flake from
+the flake input and import it into the module system. Feel free to make any
+adjustment for your setup.
+
+
 ```nix
 # home config
 
@@ -82,11 +97,40 @@ in {
 
 ## Usage
 
-Once the module is enabled, you can use it to declaratively configure Hyprland:
+Once the module is enabled, you can use it to declaratively configure Hyprland.
+Here is an example config, made to work with either the upstream Home Manager
+module, or the Flake-based Home Manager module.
 
 ```nix
 # home.nix
 {config, pkgs, ...}: {
+  # hyprland module from HM
+  wayland.windowManager.hyprland.settings = {
+    "$mod" = "SUPER";
+    bind =
+      [
+        "$mod, F, exec, firefox"
+        ", Print, exec, grimblast copy area"
+      ]
+      ++ (
+        # workspaces
+        # binds $mod + [shift +] {1..10} to [move to] workspace {1..10}
+        builtins.concatLists (builtins.genList (
+            x: let
+              ws = let
+                c = (x + 1) / 10;
+              in
+                builtins.toString (x + 1 - (c * 10));
+            in [
+              "$mod, ${ws}, workspace, ${toString (x + 1)}"
+              "$mod SHIFT, ${ws}, movetoworkspace, ${toString (x + 1)}"
+            ]
+          )
+          10)
+      );
+  };
+
+  # hyprland module from the flake
   wayland.windowManager.hyprland.extraConfig = ''
     $mod = SUPER
 
--- a/pages/Nix/Hyprland on Home Manager.md
+++ b/pages/Nix/Hyprland on Home Manager.md
@@ -2,9 +2,11 @@ For a list of available options, check the
 [Home Manager options](https://nix-community.github.io/home-manager/options.html#opt-wayland.windowManager.hyprland.enable).
 
 {{< hint title=Note >}}
-- *(Required) NixOS Module*: enables critical components needed to run Hyprland properly
-- *(Optional) Home-manager module*: lets you declaratively configure Hyprland
-{{< /hint >}}
+
+- _(Required) NixOS Module_: enables critical components needed to run Hyprland
+  properly
+- _(Optional) Home-manager module_: lets you declaratively configure Hyprland
+  {{< /hint >}}
 
 ## Installation
 
@@ -12,7 +14,8 @@ For a list of available options, check the
 
 {{< tab "Home Manager" >}}
 
-Home Manager has options for Hyprland without needing to import the Flake module.
+Home Manager has options for Hyprland without needing to import the Flake
+module.
 
 ```nix
 {
@@ -21,7 +24,16 @@ Home Manager has options for Hyprland without needing to import the Flake module
 ```
 
 {{< /tab >}}
+
 {{< tab "Flakes" >}}
+
+{{< hint >}}
+
+The flake module is merely an extension to the Home Manager downstream module.
+It is mainly used as a staging area for new options, so unless you're a tester
+you should use the downstream Home Manager module.
+
+{{< /hint >}}
 
 The following snippet of code tries to show how to bring the Hyprland flake from
 the flake input and import it into the module system. Feel free to make any
@@ -57,14 +69,22 @@ Don't forget to replace `user@hostname` with your username and hostname!
   };
 }
 ```
+
 {{< /tab >}}
 
 {{< tab "No flakes (with flake-compat)" >}}
 
+{{< hint >}}
+
+The flake module is merely an extension to the Home Manager downstream module.
+It is mainly used as a staging area for new options, so unless you're a tester
+you should use the downstream Home Manager module.
+
+{{< /hint >}}
+
 The following snippet of code tries to show how to bring the Hyprland flake from
 the flake input and import it into the module system. Feel free to make any
 adjustment for your setup.
-
 
 ```nix
 # home config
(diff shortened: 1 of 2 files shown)


haru02w (approved):
lgtm

spikespaz:
Greetings to all who come across this issue. I'm writing this to open a discussion about proceeding with migration of users to one of the new Hyprland Home-Manager modules.

To bring any readers up to speed, there are currently three HM modules for Hyprland:
- The official flake's [`nix/hm-module.nix`](https://github.com/hyprwm/Hyprland/blob/main/nix/hm-module.nix) (by @fufexan)
- The [one in HM](https://nix-community.github.io/home-manager/options.html#opt-wayland.windowManager.hyprland.enable) added recently via [#3837](https://github.com/nix-community/home-manager/pull/3837) (by @fufexan)
- And [my own](https://github.com/spikespaz/dotfiles/tree/master/hm-modules/hyprland) (by @spikespaz)

### A tale of fragmentation

To summarize the differences between the three:
1. **The one in the official flake** does little more than basic configuration. It handles the package to use (Git `master` by default), *systemd* integration, package options (*xwayland*, patches, and HiDPI), and plugin dynamically-linked libraries. It has no mechanism for generation of the config file (`$XDG_CONFIG_HOME/hypr/hyprland.conf`) and that file must either be manually created or linked in via the HM option `xdg.configFile."hypr/hyprland.config".source`.
2. **The module by @fufexan in Home Manager** uses the package in `nixpkgs-unstable` by default unless overridden in the HM input's `follows`. It expands upon (and changes some) features of the original Flake's module. Specifically, it add a the options `settings` and `extraConfig` to allow you to declaratively define your Hyprland config; this uses a [simple function](https://github.com/nix-community/home-manager/blob/master/modules/services/window-managers/hyprland.nix#L178-L200) to stringify the Nix config as `hyprland.conf`. The Nix config might look [something like this](https://github.com/nix-community/home-manager/blob/master/tests/modules/services/window-managers/hyprland/simple-config.nix#L10-L59) and the generated `hyprland.conf` looks like [this](https://github.com/nix-community/home-manager/blob/master/tests/modules/services/window-managers/hyprland/simple-config.conf). This module started out with significant issues, especially with ordering. Some things like variables and beziers must be defined before they are used (fixed in [#4263](https://github.com/nix-community/home-manager/pull/4263)). There are more caveats that are handled in the following module.
3. **My module, which is currently a [subtree of my dotfiles](https://github.com/spikespaz/dotfiles/tree/master/hm-modules/hyprland).** This is *by far* the most complete of the three modules, and provides (in my opinion) a very nice API for users to declare their Hyprland config entirely in Nix syntax without excessive interpolation and merging. The goal is to allow everything to be defined in Nix in a way that feels natural. More on this later.

### Module #3's adoption as official

In the past, @fufexan reached out to me to ask if my module could be adopted into the Hyprland flake to publicize it as official. I responded that I did not yet consider it to be ready for public adoption, as I acknowledged it is incomplete. I would consider it a prototype, albeit a fully functional one. After describing these issues, @fufexan concluded that it was "quite complex", and implied that a project of this intricacy is out of scope for the purposes of a sanctioned, official, module (I expect to be corrected). For this reason, I encouraged him to wait, and to instead test the module and provide feedback. Regardless of these issues, my module is by far the most feature-complete of the three. Because the scope of my project is wide, and I am ready and willing to accept the responsibility of maintainership, it would be a worthwhile effort to seek compromise.

My project is designed to be robust, extensible, and therefore future-proof as long as the current config format exists.

**Problems with my module:**
1. Lacking documentation, only [a few examples](https://github.com/spikespaz/dotfiles/tree/master/users/jacob/desktops/hyprland) (`config.nix`, `monitors.nix` `keybinds.nix`, and `windowrules.nix`).
	- I say this, but I did a good job with most of the the option docs. If only the task below were easy--
2. No statically generated module option documentation (potentially use [rycee/nmd](https://sr.ht/~rycee/nmd/)).
3. Missing type definitions, specifically for option `keyBinds` (uses `attrsOf anything`, should be a recursive submodule instead).
4. Unreasonable error reporting (if you get something wrong, you get a nonsensical Nix trace) for anything that isn't strictly defined (would be mitigated by solving the above issue).
5. *Some* internal functions could be simpler/optimized/eliminated.
	- **However:** the problem is inherently complicated. Details later.
6. The internal intermediate representation (essentially a node tree) is ill-defined. Currently it feels like a prototype (which it is).
	- Perhaps nodes should contain certain tokens, instead of every noun being a node.
7. Lackluster formatting options (related to #3), currently requires understanding the internals to configure (especially sorting, which is delicate and necessary).
8. Superfluous options (this is a point of contention, I want to keep--but possibly hide--the formatting options since otherwise they would just need to be hidden variables).
9. Opinionated. The file `configRenames.nix` is entirely pedantic; it serves little purpose other than to make some of the variable names (such as `col.`-prefixed ones) more compatible with the intended semantics to be expressed in the Nix config. Without this, some things would need to be quoted as attribute names, and this bothered me.
10. Config-reloading is buggy. Changes to keybinds defined with the custom syntax are not reloaded despite `reloadConfig` being `true`. This has to do with a limitation in HM's DAG, but I do have a nearly-complete work-in-progress solution in a Git stash.
11. A significant amount of time has passed since I originally drafted this module, and some important changes (namely `plugins` option) have been made to the current official module and the one in Home-Manager.
12. **Biggest blocker:** This maybe a point of contention, but I want to keep ownership and be the BDFL (coined by Linus). This entails that:
	- I approve the PRs (or Vaxry merges with mine or @fufexan's go-ahead).
	- The license is [CC-BY](https://creativecommons.org/licenses/by/4.0/) which is chosen to allow derivative works, even under a different license, but require that attribution is given to me and @fufexan (who I presume will be a core contributor).
	- I would be amenable to adjustments as long as attribution is preserved.
	- This module is *not* to be merged into Home-Manager in any official capacity. This would still be technically allowed, but unsanctioned.

**Considering #5 and #8:** I disagree with the opinion that this module ought to be simple. The problem is fundamentally complex. The problem-objective is this: to convert a structured Nix expression--consisting of attribute sets and weak dynamic types--into a node-based document language that has no specification, a weak schema, and ambiguous literal syntax. This is inherently a complicated problem that requires a complicated solution. For those who are familiar with [Serde](https://serde.rs/), that project is complicated for similar reasons. In fact, this project was inspired by that one (also see [xml-rs](https://lib.rs/crates/xml-rs), [knuffel](https://lib.rs/crates/knuffel), and [kdl](https://lib.rs/crates/kdl)).

**Why so much code?**
To expand upon the above paragraph:
- My Hyprland module for Home Manager consists of a library for serializing and formatting an attribute set into Vaxry's config language. This functionality is contained within [`configFormat.nix`](https://github.com/spikespaz/dotfiles/blob/master/hm-modules/hyprland/configFormat.nix).
- This design is heavily modeled after [`pkgs.formats`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/pkgs-lib/formats.nix) ([doc](https://github.com/NixOS/nixpkgs/blob/master/doc/functions/generators.section.md)) and I believe is the most correct way of implementing a language serializer in Nix. Currently the formatter is not exposed as any `lib` attributes, though I wish to change that to allow greater extensibility (including custom-generated files as `source`s.
- I do take come creative liberties to make the whole process conceptually simpler, otherwise it would reminiscent of a bowl of noodles. Because Vaxry's config language is essentially a node document with some additional semantics, I have chosen to model it as such. Currently (and temporarily) every section, indent, value, and variable assignment is a node of sorts, with formatting options for each type of node. 
- The target language is full of edge-cases: for example, key binds have submaps, and submaps are not a `{}`-encapsulated block, whereas in Nix they are. This decision was made to both make the code you write prettier, and to [preserve variable/section ordering](https://github.com/spikespaz/dotfiles/blob/master/hm-modules/hyprland/config.nix#L131-L169). This is because features were added to the config language ad-hoc, and (presumably) not entirely planned (no shade, Vaxry has more important things to do, like [#2415](https://github.com/hyprwm/Hyprland/issues/2415)).

### Thoughts on this PR

Now, for my review of this PR and to summarize my position. I do not think that this should be merged without further significant revision. I admit to being biased, I think that my module is best.

I do not think it wise to encourage users to migrate their configs to the technically inferior format exposed by the Home-Manager module at present, it would be premature. Proceeding would would mean that, once a consensus is reached regarding integration of my module, users would need to migrate a *second* time; users would be less inclined to do so.

As it stands, it is fairly easy to migrate a Hyprland config to a Nix config using my module. This can be done with a few replacements using regular expressions, whereas if someone were to first move to the current HM-module, and then some time later to my format, they may have written some complicated intermediate functions that conflict with the design of my module. Rectifying this would entail reversing those changes--which are probably complicated Nix expressions--and then reformatting to meet the specifications defined by my module.

This would decrease users' willingness to change their own configs, which means less people using my module, which means less feedback, which means less improvement for those who do decide to migrate. As my module is specifically *designed to be extensible*, I think that patching features and fixes into an inferior module would be a mistake, as over time it would become convoluted and unmaintainable.

Instead, I think it would be best to discuss all of the issues illustrated above and proceed with moving users over once the entire adoption process has been wholly officiated.

I look forward to further discussion and am excited at the prospect of developing this more, thank you for your patience and tolerance.

spikespaz:
For now, I encourage users to test out my module and provide feedback.
Here are some links:
- Module: https://github.com/spikespaz/dotfiles/tree/master/hm-modules/hyprland
- Examples: https://github.com/spikespaz/dotfiles/tree/master/users/jacob/desktops/hyprland
- Flake instructions: *Will update once in `README.md`.*

spikespaz:
@fufexan Do you have any comments/suggestions about the module I made? What are your requirements for using it?

spikespaz:
@fufexan [This](https://github.com/hyprwm/Hyprland/pull/870) is an entirely separate codebase, so not what you want to comment on. Open an issue on https://github.com/spikespaz/hyprland-flake and put your suggestions there. I'll have time to work on it between plane flights.

andresilva:
Between the hm-module in this repo and the one on HM repo which one should is preferred to use? It seems that the one in home-manager has been updated more recently.

spikespaz:
@andresilva That's up to you. I'm trying to get people to try out the one I made, as Fufexan has made room for it in his heart. If I know it works for people and can call it stable, I will move it to this `hyprwm` organization.

Until someone besides me is using it though, I cannot do this in good conscience.

Misterio77:
Hey all! Came across this so thought I would drop my two cents:

I think the best approach would be to adopt @fufexan's upstreamed hm module.
- Keep [hm-module.nix](https://github.com/hyprwm/Hyprland/blob/main/nix/hm-module.nix) exactly the same as HM's upstream module
- Only diverge (preferrably never diverge on interface, only in implementation) if a hyprland package absolutely requires a change in the module
    - Breaking changes should be avoiding by making sure the module follows [RFC42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md) (which fufexan's HM one already does)
    - In this case, also upstream the change to HM and merge it once `nixos-unstable` gets the updated package. Then eventually the modules will be in sync again.
- This means that people following master should usually use hyprland's provided modules, and changing between master pkg+master module and nixpkgs pkg+hm module shouldn't break their setups.

@spikespaz's module looks nice and should absolutely be an alternative for those who are interested in a more tightly integrated solution (we should link to it in the docs). But, in my opinion, it goes against the general (nixos and hm) trend of using RFC42 to make modules easier to maintain and closer to the actual config generated. Following the nixos and hm trend is, in my opinion, the way to go, as it lowers the overall complexity of setting up it (as it feels more similar to other stuff, and less of an additional DSL you have to learn).

spikespaz:
I think that [#4004](https://github.com/hyprwm/Hyprland/issues/4004) supercedes this issue.

Issue #281, opened 2023-08-01
Autostarting Apps


iynaix:
https://wiki.hyprland.org/FAQ/#how-do-i-autostart-my-favorite-apps

The given example seems overly complicated for what it's supposed to achieve.

Wouldn't the following be a much simpler way to achieve the same result?

```
exec-once=hyprctl dispatch exec '[workspace 1 silent] kitty'
exec-once=hyprctl dispatch exec '[workspace 2 silent] subl'
exec-once=hyprctl dispatch exec '[workspace 3 silent] mailspring'
exec-once=hyprctl dispatch exec '[workspace 4 silent] firefox'
```

vaxerski:
`exec-once = [workspace 1 silent] kitty` is also valid...?

iynaix:
Yep, that works as well! Used the shorter version in the PR.

PR #283, opened 2023-08-02
Simplify autostarting apps

iynaix:
Simplifies the autostarting of apps without the need for cleanup script, discussed in #281


vaxerski (approved):
k

PR #297, opened 2023-08-14
fix mistake in hyprctl activeworkspace description

waddyhussain:
was using the description for activewindow


vaxerski (approved):
right thanks

PR #299, opened 2023-08-14
Fix get-workspaces command for eww

### LINKED ISSUE #298: Documentation on eww workspace scripts may be outdated
When running the script `get-workspaces`, as shown in https://wiki.hyprland.org/Useful-Utilities/Status-Bars/#eww, the following error is shown:

```
parse error: Invalid numeric literal at line 1, column 7
jq: invalid JSON text passed to --argjson
Use jq --help for help with command-line options,
or see the jq manpage, or online docs  at https://stedolan.github.io/jq
```

Additionally, when running `get-active-workspace`, although the active workspace is eventually echoed, a similar error is shown at first:

```
parse error: Invalid numeric literal at line 1, column 7
```

These errors seem to result in the workspace widget for eww disappearing. I assume this is caused by a recent commit to hyprland (I first noticed with `8e04a80e60983f5def26bdcaea701040fea9a7ae`), since the eww and jq versions are the same as when the workspace widget was working.

MCotocel:
Closes #298

### DIFF
--- a/pages/Useful Utilities/Status-Bars.md
+++ b/pages/Useful Utilities/Status-Bars.md
@@ -103,7 +103,7 @@ Here are some example widgets that might be useful for Hyprland:
 <details>
 <summary>Workspaces widget</summary>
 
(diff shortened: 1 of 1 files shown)


MCotocel:
As of that commit, the output of `hyprctl workspaces -j` would print `adding workspaces` as the first line, resulting in the error mentioned in https://github.com/hyprwm/contrib/pull/54#issuecomment-1673602883. `sed '1d'` was meant to delete that first line, fixing the error and allowing the workspace script to function normally. I was not aware it was a bug, and assumed it was simply a deliberate change in the output. Seeing as this issue has been resolved directly in the hyprland code, I'll close this issue

PR #300, opened 2023-08-15
clarify which config to add env var

horriblename:
I've seen too many(1) people ask "where do I put this"


vaxerski (approved):
true, people are idiots

PR #305, opened 2023-08-17
Animations: document slidefade and slidefadevert styles for workspaces

end-4:
https://github.com/hyprwm/Hyprland/pull/3008


vaxerski (approved):
g

rehanzo:
The provided example doesn't work, probably because it seems these are workspace animations, and not window animations?

end-4:
ahhh right i was stupid

PR #306, opened 2023-08-18
Animations: fix slidefade example

end-4:
slidefade/slidefadevert are for workspaces not windows


end-4:
fixes mistake of #305

vaxerski (approved):
how did I not notice

PR #312, opened 2023-08-23
Add waypaper to wallpaper tools

anufrievroman:
Adding a GUI tool called Waypaper to the list of wallpaper tools


vaxerski (approved):
self-plug innit

PR #316, opened 2023-08-28
Copy empty string into primary clipboard instead of clearing selection

ti-mo:
I'm using this wl-paste/wl-copy setup to disable middle click, but I've noticed `wl-copy -pc` also clears the selected text in GTK UI elements, which prevents the user from selecting text altogether.

I've found a similar report at https://bugs.kde.org/show_bug.cgi?id=461498. Not sure where the issue lies exactly, but `wl-copy -p ''` accomplishes the same (clearing the primary buffer) without also clearing the selected text.


vaxerski (approved):
ok

PR #318, opened 2023-08-30
Doc, window rules: Mention what pin does.

LiamKearn:
The description as it is can be implied from the name, it's pretty useless without the context. I've just copied that over from dispatchers.


vaxerski (approved):
now watch me whip 👊
now watch me nae nae 👋

PR #321, opened 2023-09-03
Added a mention to a slackbuild under a new "Slackware" tab.

tmheath:
Instead of going into detail, I felt that it might be best only to link, as even though Slackware is not likely to update very soon the slackbuild itself is lagging behind the most current version by a lot, I did not feel comfortable that the instructions from the build would not deviate. I can provide the detailed instructions if you instead feel that this is incorrect.


vaxerski (approved):
thanks :)

PR #323, opened 2023-09-03
Remove warnings about Spotify not respecting window rules

shasherazi:
There are some warnings about Spotify not respecting window rules in the docs.

https://github.com/hyprwm/hyprland-wiki/blob/b8de59bf08eec5f1d8103e16744b749ff4315b93/pages/Useful%20Utilities/App-Clients.md?plain=1#L16-L17

As it turns out, Spotify now follows the window rules, e.g., opening in its designated workspace, etc. So, these lines are not needed anymore.


vaxerski (approved):
ooga booga

PR #332, opened 2023-09-09
Update Installation.md

shadowmax31:
Needs gcc 13+, I don't know if it changed anything for clang.

https://github.com/hyprwm/Hyprland/pull/3186


fufexan:
Are you sure it works with `13.0.0`? I've only tried building with `13.2.0`.

vaxerski (approved):
ok, but clang is also wrong now lel

shadowmax31:
> Are you sure it works with `13.0.0`? I've only tried building with `13.2.0`.

I also only tried with 13.2.0, but I went off of vaxry's comment in the PR

vaxerski:
it's 13.0

PR #337, opened 2023-09-11
Alterzorder

alaricljs:
for https://github.com/hyprwm/Hyprland/pull/3217


vaxerski (approved):
top

Aqa-Ib:
It would be good if we add in the description that this only works for two floating windows, but not for a tiled and a floated window, if I am right. Thanks for the work.

PR #343, opened 2023-09-13
Add note regarding screen sharing with 10 bit enabled

filahf:
10 bit and screen sharing is a known issue:

https://github.com/hyprwm/xdg-desktop-portal-hyprland/issues/24
https://github.com/emersion/xdg-desktop-portal-wlr/issues/239

### DIFF
--- a/pages/Configuring/Monitors.md
+++ b/pages/Configuring/Monitors.md
@@ -157,9 +157,9 @@ end of the monitor rule, e.g.:
 ```ini
 monitor=eDP-1,2880x1800@90,0x0,1,bitdepth,10
 ```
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
it's not a problem with xdp, it's a problem with whatever app is capturing the buffers if it doesn't support 10-bit ones. We shouldn't explicite say xdp

filahf:
> it's not a problem with xdp, it's a problem with whatever app is capturing the buffers if it doesn't support 10-bit ones. We shouldn't explicite say xdp

Ah, I see. I'll reword that bit

vaxerski (approved):
g

PR #359, opened 2023-09-29
Nvidia: Add tips to fixing suspend issues

tchofy:
First PR ever, so here we go:

Nvidia's proprietary drivers include some systemd services to aid with suspend/hibernate and they aren't enabled by default in any of the distros I've seen, and their userspace driver only restores essential memory on wakeup, leading to graphical corruption and crashes, unless their driver is loaded with a specific parameter.

### DIFF
--- a/pages/Nvidia/_index.md
+++ b/pages/Nvidia/_index.md
@@ -104,3 +104,11 @@ options nvidia NVreg_RegistryDwords="PowerMizerEnable=0x1; PerfLevelSrc=0x2222; 
 ```
 
 Reboot your computer and it should be working.
+
+## Fixing suspend/wakeup issues
+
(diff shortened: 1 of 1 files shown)


fufexan (approved):
If it works on Arch, LGTM.

tchofy:
Should be the same on debian and fedora based distros as well except for driver package name.

From my brief experience on Nixos a couple months ago, the equivalents for these are:
 
```
boot.kernelParams = [ "nvidia.NVreg_PreserveVideoMemoryAllocations=1" ];
...
#false by default
hardware.nvidia.powerManagement.enabled = true

#false by default, but good to add in case of upstream changes
hardware.nvidia.open = false 
```
Should I include these in the nvidia page as well?

fufexan:
Sure, why not.

tchofy:
Done

fufexan (approved):

PR #360, opened 2023-09-30
added wiki entry for my PR

DaniD3v:
https://github.com/hyprwm/Hyprland/pull/3459


vaxerski (commented):
please don't `<br/>` we don't want brazil in our md

vaxerski:
also come to think of it, but this is more to the hyprland MR, where is the option to use a random anime wp?

Maybe make it -2?

vaxerski (commented):
nice, just mention that setting force to 0 will use the non-anime wallpaper, many people might not pick up on that.

vaxerski (approved):

PR #361, opened 2023-09-30
Keep the Nouveau driver in mind.

Aqa-Ib:
Mention the Nouveau driver. In my case, it is far better than nvidia or nvidia-open. No problems at all with it using multiple monitors.


vaxerski (commented):
I'd at least mention performance and feature set might be worse.

vaxerski (approved):
g

PR #367, opened 2023-10-09
Add persistent workspace rule

hariseldon78:
This is to document the rule added in https://github.com/hyprwm/Hyprland/pull/3530.


pages/Configuring/Workspace-Rules.md
```diff
@@ -25,6 +25,7 @@ workspace=DP-1,3,rounding:false,bordersize:0
 | ---- | ----------- | ---- |
 | monitor:[m] | Binds a workspace to a monitor See [syntax](#syntax) and [Monitors](../Monitors).| string |
 | default:[b] | Whether this workspace should be the default workspace for the given monitor | bool |
+| persistent:[b] | Keep this workspace alive even if empty and inactive | bool |
```
vaxerski, line 28: we generally add to the bottom of a list unless something something

hariseldon78: ok but all the other flags refer to window properties, that's why i grouped it with default. But noo problem, i'll change it

vaxerski (approved):
g

PR #369, opened 2023-10-11
Remove unnecessary optimization from FAQ

Tasztalos69:
The `new_optimizations` field was incorrectly nested on the FAQ page.


vaxerski (commented):
I'd rather remove the line completely tbh as it's on by default and should not be touched

vaxerski (approved):
true, on by default

PR #370, opened 2023-10-12
Update Master-Tutorial.md

Blubberland:
Beeing more specific about forcing electron apps to use wayland.

Bugged me a bit, until I found out about it.


vaxerski (commented):
it's not required, at least for me. Maybe note both cases?

stickyburn:
I have never needed it either :thinking: 

Also @vaxerski , vscode does work with `code-flags.conf`, should I add that too?

vaxerski:
u can

Blubberland:
Sorry for not taking part of the conversation anymore.
I don't have access to the device with this specific configuration and versions etc and I didn't git it.
I had this issue with bitwarden. Maybe it's been solved since then. I wanted to look further into the whole topic, but I didn't know where to start.

Maybe the problem was a very old electron version for bitwarden. I'm not sure how to start troubleshooting stuff like this.
I just found the solution somewhere, tried it, and it worked for me.
So I thought, it might be a good idea to give a hint in the docs. 

It might take some time for me to get access to the mentioned device again, but I might be able to try to reproduce it on another device.

PR #382, opened 2023-11-03
Added FAQ item about TLP causing stutter on Intel

tobimai:
The default TLP settings caused a lot of stutter and FPS drops when animations ran. It dropped to under 30 FPS from 60 on all animations. 

This was solved by increasing the minimum GPU frequency specified in TLP. I did some tests, it doesn´t seem to have any worse power consumption as the GPU goes into some deeper power-saving when idle anyway, but animations now run perfectly smooth at 60 FPS (which is the refresh rate of my Panel). 

And as it took me a while to figure it out I thought adding it to the FAQ, which I read like 5 times would be worth it.

### DIFF
--- a/pages/FAQ/_index.md
+++ b/pages/FAQ/_index.md
@@ -291,3 +291,6 @@ set `MOD` and `KEY` to desired values.
 By pressing the selected combo you will enter a mode where hyprland ignores your keybinds and passes them on to the vm.
 
 Then, pressing `SUPER + Escape` will leave that mode.
+
+# Low FPS/stutter/FPS drops on intel iGPU with TLP (mainly laptops)
+The TLP defaults are rather agressive, setting `INTEL_GPU_MIN_FREQ_ON_AC` and/or `INTEL_GPU_MIN_FREQ_ON_BAT` in `/etc/tlp.conf` to something slightly higer (e.g. to 500 from 300) will reduce stutter significantly or in the best case remove it completly.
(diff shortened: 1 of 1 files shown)


pages/FAQ/_index.md
```diff
@@ -291,3 +291,6 @@ set `MOD` and `KEY` to desired values.
 By pressing the selected combo you will enter a mode where hyprland ignores your keybinds and passes them on to the vm.
 
 Then, pressing `SUPER + Escape` will leave that mode.
+
+# Low FPS/stutter/FPS drops on intel iGPU with TLP (mainly laptops)
+The TLP defaults are rather agressive, setting `INTEL_GPU_MIN_FREQ_ON_AC` and/or `INTEL_GPU_MIN_FREQ_ON_BAT` in `/etc/tlp.conf` to something slightly higer (e.g. to 500 from 300) will reduce stutter significantly or, in the best case, remove it completly.
```
fufexan, line 296:
```suggestion
# Low FPS/stutter/FPS drops on Intel iGPU with TLP (mainly laptops)

The TLP defaults are rather aggressive, setting `INTEL_GPU_MIN_FREQ_ON_AC` and/or `INTEL_GPU_MIN_FREQ_ON_BAT` in `/etc/tlp.conf` to something slightly higher (e.g. to 500 from 300) will reduce stutter significantly or, in the best case, remove it completely.
```

fufexan (approved):

PR #383, opened 2023-11-04
Hyprland on NixOS: Fix link to Home Manager page

TenEXP:
The link currently leads to a directory named "Hyprland on Home Manager" instead of the markdown file named as such. Added .md to the link to fix this.


TenEXP:
Mistakes were made. Testing was foregone.

PR #384, opened 2023-11-04
Expand environment variables section.

Caellian:
Add Java AWT to toolkits.
Add application specific variables section for accumulating all the PLZ_WAYLAND env flags.

### DIFF
--- a/pages/Configuring/Environment-variables.md
+++ b/pages/Configuring/Environment-variables.md
@@ -29,7 +29,6 @@ Linux distros.
   compatibility issues
 - `CLUTTER_BACKEND=wayland` - Clutter package already has wayland enabled, this variable will force Clutter applications
   to try and use the Wayland backend
-- `_JAVA_AWT_WM_NONREPARENTING=1` - Make Java AWT accept redraw/input events without waiting for a reparenting event.
 
 # XDG Specifications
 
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
nonreparenting and moz are automatically set by hl. Worth noting...

Caellian:
> nonreparenting and moz are automatically set by hl

I moved them to a separate section so users know they don't require setting. Let me know whether that's fine, should be removed altogether, or just a comment next to the items would've been better.

If the separate section is kept, I guess I'll go through sources to add any other variables that are configured - just the ones that interact with other programs (i.e. not set by all WMs)

fufexan:
You may want to also add `SDL_VIDEODRIVER=wayland`, `QT_QPA_PLATFORM=wayland` and `XDG_SESSION_TYPE=wayland`.

vaxerski:
already set also includes XDG_SESSION_TYPE and XDG_CURRENT_DESKTOP for sure.

fufexan:
I didn't see them when grepping through the source, my bad.

Caellian:
Not sure whether you want "already set" section to be included @vaxerski. If I add it, you'll have to keep it up to date so lmk. IDK whether env vars are something you add/change often so I don't want to unnecessarily pile onto to the maintenance burden.

vaxerski:
they arent changed often

Caellian:
I don't think these are needed anymore for the most part.

PR #394, opened 2023-11-12
For discussion, NO MERGE : Dispatcher v2 API 1st proposition

johnr14:
The current API of dispatcher is not optimal, it's not uniform, uses too many types of different actions and has poor predictability for function's name.

Left unchanged, it is a irritant to new user and will get worse with time if the complexity of hyprland keeps growing in a far future. Those changes can be postponed, but having a API revision after release 1.0 wouldn't be the best.

As no names where kept and thus no collision with previous API occurs, *this could be implemented without breaking anything*.

Old API could be discouraged and eventually be deprecated and removed after a notice and appropriate elapsed time.

The term `window` is deprecated in favor of `client` as Wayland namely use clients. However Wayland does interchange the term `window` and `client` a lot in it's documentation but it says this about it's [architecture](https://wayland.freedesktop.org/architecture.html) : "is how **clients** actually render under wayland", "the rendering happens in the **client**, and the **client** just sends a request to the compositor to indicate the region that was updated ", "the compositor receives generic Wayland buffers from the **clients**". _Could revert to windows if maintainers or users strongly wish for it_.

This proposition is an attempt to consolidate the wording used in the dispatcher's functions name with a naming convention that :
```
    1. would help understand what a function does without having to read it's description.
    2. be uniform
         2.1 for the type of action it performs
         2.2 for the subject of the action
         2.3 for the target of the action
         2.4 for the passing of arguments, mandatory and optional
```

All names are in a format of : `nounVerbVariation`.

The number of dispatchers action verbs went down from **22** to **9**.

Lots of param where defined. Params use the `opt:param` syntax, may or may not be kept depending on difficulty to implement. 

A few examples are added for groups and special workspace.

Discussed in [#3805 ](https://github.com/hyprwm/Hyprland/discussions/3805)

First proposition, please check for : 
1. if it makes any sens to change the API
2. coherence and logic
3. omissions
4. ease of use
5. efficiency
6. orthography and syntax after first 3 :) 

This is in **no way implemented in bind or hyprctl** and is used to illustrate how the API could be redesigned. 

Thanks for your time in pondering on how to make hyprland even greater and please add your comments. I did try to finish in a rush and may have made a few "oops". I'll try to squash the request, hope it works.

### DIFF
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -30,6 +30,7 @@ Dispatchers names are not case sensitive and uppercase is only use for easier re
 | `orgroup` | When moving a client, performs different directionnal move : if in a group, will move out of it; if not in group, will move in; if neither, will just move the client. |
 | `out` | Move out of a group but keeps group flag active. |
 | `prev` or `next` | Select previous or next element in a sequence. |
+| `reset` | Used to exit a `Submap` |
 | `resizeparams` | relative pixel delta vec2 (e.g. `10 -10`), optionally a percentage of the window size (e.g. `20 25%`) or `exact` followed by an exact vec2 (e.g. `exact 1280 720`), optionally a percentage of the screen size (e.g. `exact 50% 50%`) |
 | `stack` | Move to the stack `top` or `bottom`. |
 | `submapname` | A name for a submap. |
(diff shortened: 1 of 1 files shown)


johnr14:
Please view as a [wiki-page](https://github.com/hyprwm/hyprland-wiki/blob/e57cf6668585cafea33af644598183a7a8030966/pages/Configuring/Dispatchers.md) for now. Thanks

TuxMC-sys:
> Please view as a [wiki-page](https://github.com/hyprwm/hyprland-wiki/blob/e57cf6668585cafea33af644598183a7a8030966/pages/Configuring/Dispatchers.md) for now. Thanks
> 
> If you know how I could squash this, please tell.

Is [this](https://gist.github.com/lpranam/4ae996b0a4bc37448dc80356efbca7fa) what you're looking for?

PR #403, opened 2023-11-25
add window rule case sensitivity disclaimer

jacekpoz:
my motivation for this is [this discussion](https://github.com/hyprwm/Hyprland/discussions/3903)


vaxerski (approved):
obvious, as it's regex, but ok

PR #404, opened 2023-11-25
nav: make "Configuring Hyprland" a link and remove "Start Configuring"

rszyma:
Remove "Start Configuring" link in navbar and make "Configuring Hyprland" point to the page that previously "Start Configuring" pointed to.

This is already done for "Useful Utilities" category, and IMO should be done for other categories.

Picture explanation of the changes:

image


vaxerski:
this would require changes in the backend, the sidebar is not exposed

rszyma:
> this would require changes in the backend, the sidebar is not exposed
> this is already done for "Useful Utilities" category

rszyma:
> also, I am not a fan

why? some navbar links being clickable and other not makes experience of browing wiki inconsistent and when there multiple inconsistencies like that, it will be annoying to user

rszyma:
> this would require changes in the backend

ah, I see, you mean the "Configure" button in main repo. It's whatever I'll close this one.

PR #405, opened 2023-11-26
Perfomance category suggestion

RadovanSk:
After finding a solution to this problem I had https://github.com/hyprwm/Hyprland/issues/3959 I thought it would be a good idea to document it so other people know what to do if they stumble upon the same problem. Not only that, I also believe that it would be great to have such category in the hyprland wiki where tricks and known solutions could be documented for people who seek to get more performance out of hyprland.

For now, there is only fix for fractional scaling from the mentioned issue. Though I believe that this page would be very useful when more stuff gets added to it.

Any suggestions or recommendations would be appreciated!

### DIFF
--- a/pages/Configuring/Perfomance.md
+++ b/pages/Configuring/Perfomance.md
@@ -3,3 +3,23 @@ This page documents known tricks and fixes to boost perfomance if by any chance 
 # Fractional scaling
 
 Wayland fractional scaling is a lot better than before, but it is not perfect. Some applications do not support it yet or the support is experimental at best. If you have problems with your graphics card having high usage and hyprland feeling laggy then try setting the scaling to integer numbers such as `1` or `2` like in this example `monitor=,preferred,auto,2`.
+
+# Low FPS/stutter/FPS drops on Intel iGPU with TLP (mainly laptops)
+
+The TLP defaults are rather aggressive, setting `INTEL_GPU_MIN_FREQ_ON_AC` and/or `INTEL_GPU_MIN_FREQ_ON_BAT` in `/etc/tlp.conf` to something slightly higher (e.g. to 500 from 300) will reduce stutter significantly or, in the best case, remove it completely.
+
+# How do I make Hyprland draw as little power as possible on my laptop?
+
+I assume you already have `damage_tracking` on full. If you don't, change it.
+It's heavily advised to use `full` regardless of anything.
+
+**_Useful Optimizations_**:
+
+* `decoration:blur = false` and `decoration:drop_shadow = false` to disable
+   fancy but battery hungry effects.
+
+* `misc:vfr = true`, since it'll lower the amount of sent frames when nothing is happening on-screen.
+
+# My games work poorly, especially proton ones
+
+Use `gamescope`, tends to fix any and all issues with wayland/Hyprland.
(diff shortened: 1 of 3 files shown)


vaxerski:
why not slap it into FAQ like we do with battery-drain related stuff?

RadovanSk:
> why not slap it into FAQ like we do with battery-drain related stuff?

I was thinking that it would suit configuration more because in the end it would be basically telling you how to configure something to boost performance. However, if you believe that it would suit FAQ more then I can change that.

RadovanSk:
> > why not slap it into FAQ like we do with battery-drain related stuff?
> 
> I was thinking that it would suit configuration more because in the end it would be basically telling you how to configure something to boost performance. However, if you believe that it would suit FAQ more then I can change that.

Also to add. There may be a lot of ways to boost performance and it might be too big for FAQ. That's also why I thought it would fit more into configuring.

RadovanSk:
> fair enough, then move them

Done, what do you think?

pages/Configuring/Perfomance.md
```diff
@@ -0,0 +1,25 @@
...
+It's heavily advised to use `full` regardless of anything.
+
+**_Useful Optimizations_**:
+
+* `decoration:blur = false` and `decoration:drop_shadow = false` to disable
+   fancy but battery hungry effects.
+
+* `misc:vfr = true`, since it'll lower the amount of sent frames when nothing is happening on-screen.
+
+# My games work poorly, especially proton ones
```
vaxerski, line 23: this is not just about performance, I'd either move it back to faq or have it in both. It's also about bugs.

pages/Configuring/Perfomance.md
```diff
@@ -0,0 +1,25 @@
...
+
+Wayland fractional scaling is a lot better than before, but it is not perfect. Some applications do not support it yet or the support is experimental at best. If you have problems with your graphics card having high usage and hyprland feeling laggy then try setting the scaling to integer numbers such as `1` or `2` like in this example `monitor=,preferred,auto,2`.
+
+# Low FPS/stutter/FPS drops on Intel iGPU with TLP (mainly laptops)
+
+The TLP defaults are rather aggressive, setting `INTEL_GPU_MIN_FREQ_ON_AC` and/or `INTEL_GPU_MIN_FREQ_ON_BAT` in `/etc/tlp.conf` to something slightly higher (e.g. to 500 from 300) will reduce stutter significantly or, in the best case, remove it completely.
+
+# How do I make Hyprland draw as little power as possible on my laptop?
+
+I assume you already have `damage_tracking` on full. If you don't, change it.
```
vaxerski, line 13: tbh this can be removed, the option is hidden

RadovanSk: Should it be added back to FAQ or deleted altogether?

pages/Configuring/Perfomance.md
RadovanSk: I'll put it in both then.

RadovanSk:
Alright I made the changes. Does it look good or did I make some mistake?

RadovanSk:
@vaxerski So what do you think? Is it fine now or should I change something.

vaxerski:
sorry I am just really busy this week and missed this

vaxerski (approved):

PR #408, opened 2023-12-05
binds: Remove inaccurate comment about case and clarify xkb keysym usage

alerque:
Tangentially related to https://github.com/hyprwm/Hyprland/issues/4045. I was researching if anything had changed with regard to bindings and ran across this bit in the docs that is both unclear and seemingly inaccurate. This is my attempt to clarify what keysym names to use as well as fix the inaccuracy.


vaxerski (approved):
this is correct, the case is ignored. Even sPoNgEbOb cAsE should work:

https://github.com/hyprwm/Hyprland/blob/83248b6936c00ddef6a39fa31a37949ca1a34ffe/src/managers/KeybindManager.cpp#L456-L457

PR #412, opened 2023-12-07
Add tomlplusplus as a dependency on Ubuntu and Arch Linux

Evangelospro:
Tomlplusplus is required for hyprland latest plugin manager hyprpm as already noted here: https://github.com/hyprwm/Hyprland/commit/620031fd293b10e6cad4eee448eb8201f97477f4


vaxerski (approved):
thanks :)

PR #424, opened 2023-12-24
Fix grammar in mouse cursor change question

zootedb0t:


fufexan:
I don't know if you noticed but there are many such deliberate spelling mistakes in the FAQ. Vaxry made them intentionally, although it may be time to correct them. @vaxerski thoughts?

vaxerski:
that's valid use of words in australian or irish english

"I went to me mum"

zootedb0t:
Cool

PR #426, opened 2023-12-27
Fix typo in pages/Plugins/Development/Getting-Started.md

Jero075:
Just a small fix for a typo on line 105 ("hyprlandd.conf" instead of "hyprland.conf") to avoid any confusion. (not very important)


fufexan:
It's the debug build, and it uses the `hyprlandd.conf` file specifically. It's not a typo.

PR #431, opened 2024-01-05
Fix Gentoo XDPH installation instructions

seven-mile:
In #271, XDPH was marked as "in the main tree". But in fact XDPH is only available in GURU.

[XDPH in GURU repo](https://github.com/gentoo/guru/tree/master/gui-libs/xdg-desktop-portal-hyprland).


vaxerski (approved):
thanks!

PR #433, opened 2024-01-07
Mention plugin availability in the GURU Gentoo overlay

foopsss:
I've noticed that some plugins can be installed from GURU now, so I added that information.

### DIFF
--- a/pages/Getting Started/Installation.md
+++ b/pages/Getting Started/Installation.md
@@ -82,7 +82,7 @@ The hyprland package is available in the main tree:
 emerge --ask gui-wm/hyprland
 ```
 
-Additional packages like xdg-desktop-portal-hyprland, hyprland-plugins, hyprpaper and hyprpicker are available in the [GURU](https://wiki.gentoo.org/wiki/Project:GURU) overlay. Community-contributed scripts are also available in GURU as part of the hyprland-contrib package.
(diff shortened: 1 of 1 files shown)


vaxerski:
I would advise against, we have `hyprpm`.

foopsss:
I'm aware of `hyprpm`, but, from my short experience on Gentoo it seems that people here also like having the option to install plug-ins for apps using Portage. A good example of this is [Emacs](https://wiki.gentoo.org/wiki/GNU_Emacs#Additional_elisp_packages). Moreover, I think that not every plug-in has updated its documentation to add a `hyprpm` installation method yet, and I don't know if instructions are generic enough that they can be used for every plug-in, since I don't use plug-ins.

However, I don't intend to go against your wishes, so I can just close the MR if desired and add this to the Gentoo wiki page for Hyprland instead.

vaxerski:
well the problem with hyprland plugins is that plugins have to be compiled against the exact version of hyprland you're running. If you update your hyprland but don't recompile plugins, they stop working (or even worse, could crash hyprland)

vaxerski:
I added hyprpm for a reason. All plugins have either a hyprload or a hyprpm manifest, so this shouldnt be an issue. Instructions are generic.

PR #434, opened 2024-01-09
remove waybar wlr/workspaces patch

MightyPlaza:
removes reference to `wlr/workspaces` patch since `hyprland/workspace` should probably be used, and if I remember it correctly the `wlr/workspaces` protocol doesn't even work anymore


vaxerski (approved):
right

PR #435, opened 2024-01-10
fix bug caused by xwaylandvideobridge example configuration

Aqa-Ib:
The nofocus rule causes a disgusting behaviour when launching a program with rofi in the workspace where xwaylandvideobridge silently resides, causing it to launch in the last used workspace with an open window instead.


Aqa-Ib:
Added two more lines to get the window out of the way. Not related to the initial issue though.

vaxerski (approved):
k

PR #452, opened 2024-01-31
Add hyprshade to "other" useful utilities

loqusion:
I'm totally unbiased here 🙂


vaxerski:
I would add the color correction is only for nvidia users because nvidia doesn't support the standard way of implementing color temperature shifting

loqusion:
done :)

fufexan (approved):

vaxerski (approved):
cock

PR #459, opened 2024-02-09

2hexed:
Added a quick little hack for toggling touchpad on the fly

### DIFF
--- a/pages/Configuring/Uncommon-tips-&-tricks.md
+++ b/pages/Configuring/Uncommon-tips-&-tricks.md
@@ -1,3 +1,13 @@
+# Touchpad state toggle
+
+If one wishes to toggle the state of the touchpad on the fly, they may do so via:
+
+```ini
+bind=,XF86TouchpadToggle,exec,hyprctl keyword "device:TOUCHPAD-IDENTIFIER:enabled" true|false
+```
+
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
IMO not clear enough that `true|false` is not a valid thing and one should be picked

twohexed:
> IMO not clear enough that `true|false` is not a valid thing and one should be picked

Honestly, the `true|false` thing was a hit or miss for me as I'm new to hyprland. Since most of other WMs make use of the pipe operator (`|`) in order to do perform one thing or the other (left hand or right hand) according to the number of times the key combination is encountered (ie. `EXECUTE THIS FIRST TIME|EXECUTE THIS THE SECOND TIME`).

However, I did stumble upon something I cannot say for sure if this is a bug or not because I tested this thing in a hurry but, Say for example you're using the exact same keybinding I provided and you disable your Touchpad with that keybind on the fly. Now, if you open up a terminal and manually try to enable the device with `hyprctl keyword blabla` it does output `ok` but does not actually enable your Touchpad.

Still learning about Hyprland though, do correct me wherever needed, I'm just happy to contribute to Hypr.

vaxerski:
hyprland doesn't do that. This can be doable with a script.

vaxerski:
I don't think it does, unless your shell does this.

PR #465, opened 2024-02-11
Added info about non-systemd distros for pipewire

o0nd7ots:
It will help other people running non-systemd distros


vaxerski (approved):
lgtm

PR #466, opened 2024-02-11
Reformatted sentences

o0nd7ots:
They are better now I think


vaxerski:
I dont like how it's now seemingly less "important"

vaxerski (approved):
alr

PR #467, opened 2024-02-11
Added Emacs as a lancher suggestion

o0nd7ots:
Credit goes to DistroTube


vaxerski:
this is horrible, why?

o0nd7ots:
do you mean the idea in general or my commit specifically?

vaxerski:
idea. Why use emacs?

o0nd7ots:
There are people that use Emacs for **a lot** of stuff. Emacs can even be a full fletched [window manager](https://github.com/emacs-exwm/exwm).
It's no wonder that someone decided that they want to use Emacs as a run launcher.

I think adding it as a wiki entry, not necessarily with code, will expose this option to more people.
I see no reason for it not to be mentioned there -- from what I have noticed, folks not using full DEs like to keep their stuff minimal.
Letting them know that they can replace yet another app with one that they might already have installed is at least interesting.
Since version 28.1 lisp files can be cimpiled to native code, so users who will choose to use this will have a similar performance with less memory and storage usage.

This solution is obviously geared towards who have an Emacs server running, so they use it a lot.
For those people this entry will be especially tempting.

o0nd7ots:
It's actually GPLed:
https://github.com/SebastienWae/app-launcher

vaxerski:
and this wiki is not GPLed, so we can't use it.

o0nd7ots:
If you say so.

I think this wouldn't be a problem if we left a note that it's GPLed, since we are "quoting" it.
Nevertheless the site looks cleaner without it.

vaxerski (approved):
lgtm

o0nd7ots:
Thanks for being patient!

vaxerski:
yours truly

急いては事を仕損じる

PR #470, opened 2024-02-13
Example Volume Media Key Binds

ethangamma24:
Added example bindings for volume media keys in pages/Configuring/Binds.md in an Examples section at the bottom


vaxerski (commented):
raise and lower could use the `e` flag too.

ethangamma24:
Good call! I'll add that in :)

PR #479, opened 2024-02-19

yassineibr:
Fix the gamemode.sh script


fufexan:
In both cases (old & new change), it prints the same thing for me:
```
0
true
```

yassineibr:
I think that the docs take the second line with `NR==2`, which returns true and not 0 or 1.

fufexan:
Oh yeah, my bad. I was using `NR=2` instead of `NR==2`. Classic mistake.

fufexan (approved):

PR #483, opened 2024-02-20
Fix input field variable name

asiantuntija:
Placeholder variable uses underscore instead of dash, minor fix.

edit: also first time doing a pull request via web interface, don't know why it adds the second line


vaxerski (approved):
missed shift

PR #485, opened 2024-02-20
hyprlock: mention dot's config options

bvr-yr:
ref https://github.com/hyprwm/hyprlock/pull/24


bvr-yr:
probably we need to make variables section or each widget because it grows quickly enough

PR #491, opened 2024-02-24
Gentoo installation instructions for hypridle and hyprlock

foopsss:
As per the title, this PR mentions that hypridle and hyprlock are available on GURU.


foopsss:
Not related, but I also found out that hyprlang is in [the main tree](https://packages.gentoo.org/packages/dev-libs/hyprlang) as well, although I don't know if it's worth mentioning in the installation instructions.

vaxerski (approved):
oke

PR #496, opened 2024-03-01
add full hypridle example work with hyprlock.

sentakuhm:
added full hypridle example work with hyprlock for better understanding.

### DIFF
--- a/pages/Hypr Ecosystem/hypridle.md
+++ b/pages/Hypr Ecosystem/hypridle.md
@@ -41,14 +41,6 @@ You can define as many listeners as you want.
 Full hypridle example with hyprlock
 
 ```ini
-general {
-    lock_cmd = notify-send "lock!"          # dbus/sysd lock command (loginctl lock-session) 
-    unlock_cmd = notify-send "unlock!"      # same as above, but unlock
-    before_sleep_cmd = notify-send "Zzz"    # command ran before sleep
-    after_sleep_cmd = notify-send "Awake!"  # command ran after sleep
-    ignore_dbus_inhibit = false             # whether to ignore dbus-sent idle-inhibit requests (used by e.g. firefox or steam)
-}
-
 listener {
     timeout = 300                           # 5min
     on-timeout = hyprlock                   # lock screen when timeout has passed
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
since this is meant as an example, the notify-send stuff in general {} can be removed for better readability

sentakuhm:
> since this is meant as an example, the notify-send stuff in general {} can be removed for better readability

done.

vaxerski (approved):
lgtm, thanks!

dpredovic:
The example could be expanded to avoid questions/issues. It doesn't lock before suspend, connects to dbus or shows how to lock/suspend manually.

I'd suggest something like:
```
general {
    lock_cmd = pidof hyprlock || hyprlock # avoid starting multiple hyprlock instances
    before_sleep_cmd = loginctl lock-session # lock before suspend
    after_sleep_cmd = hyprctl dispatch dpms on # to avoid having to press a key twice to turn on the display
}

# turn off keyboard backlight
#listener { 
#    timeout = 30 
#    on-timeout = brightnessctl -sd rgb:kbd_backlight set 0 
#    on-resume = brightnessctl -rd rgb:kbd_backlight
#}

# set monitor backlight to minimum
listener {
    timeout = 60
    on-timeout = brightnessctl -s set 0 
    on-resume = brightnessctl -r
}

# lock
listener {
    timeout = 120 
    on-timeout = loginctl lock-session 
}

# turn display off
listener {
    timeout = 150
    on-timeout = hyprctl dispatch dpms off
}

# suspend
listener {
    timeout = 180
    on-timeout = systemctl suspend
}
```

You also most probably want to have something like this in hyprland.conf:
```
exec-once = hypridle
...
misc {
  mouse_move_enables_dpms = true
  key_press_enables_dpms = true
  ...
}
...
bind = $mainMod, L, exec, loginctl lock-session 
bind = $mainMod ALT, L, exec, systemctl suspend 
```

vaxerski:
that'd be better off for the hypridle page

dpredovic:
Yes, but the documentation is very... Abstract? "notify-send" for every option doesn't give users much of a direction of how is it supposed to be integrated with rest of the system. 

If you look at github dotfiles repos containing hypridle, you will hardly find one where it's integrated "properly", so I'd say that even powerusers have troubles with full config. I needed hours to come up with a config that works in all edge cases. :)

sentakuhm:
@vaxerski so, are we going to add @dpredovic's example?.

vaxerski:
idk feel free to mr

PR #511, opened 2024-03-07
Fix bash shebangs for NixOS users

jayrahdevore:
I know it's a bit of a nit pick, but copy-paste won't work when trying the examples for eww bar in NixOS.  There are a couple of places in the documentation where `/bin/bash` is referenced (this doesn't exist on NixOS).  For compatibility reasons, I've changed references to `/bin/bash` to `/usr/bin/env bash`.  `/bin/sh` is intact since it isn't an issue.


fufexan (approved):
LGTM, thanks!

PR #531, opened 2024-03-14
Add pinentry window rule example

haasn:
I think more people will find this one useful, as it's incredibly common (and frustrating) to lose focus while typing a password into a pinentry prompt, accidentally leaking parts of the passphrase in plaintext into whatever terminal / program I had previously focused.

(Strangely, this seems to happen all the time for me even with `follow_mouse = 2`..)


vaxerski (approved):
sure, whatever that is

PR #532, opened 2024-03-14
Remove outdated information

AngeloThys:
Remove a hint that suggested the name of the device be written as `device:<device-name>`.


vaxerski (approved):
oops, missed that. Thanks

PR #535, opened 2024-03-15
add walker app launcher

abenz1267:


vaxerski (approved):
self-promo at its finest

PR #540, opened 2024-03-17
Fix table formatting in Workspace Rules

rdvm:
Was just reviewing options in the wiki and noticed this table was borked.


fufexan (approved):
Oops, missed it. Thanks!

PR #559, opened 2024-03-27
Show how you can set keyboard layout variants per layout

Marcuss2:
Added an example how to set keyboard variant.


vaxerski (approved):
lgtm

PR #563, opened 2024-03-30
Update document mistake with initialClass with added warnings

Trap101:
The doc before was wrong on initialclass and is quite confusing, changed the doc from initialClass to initalclass and added warning to make sure this is aware by the reader.


vaxerski (approved):
k

PR #568, opened 2024-03-31
Add text to 3 empty pages

k26pl:
It ain't much, but it's honest work
edits to add more content or make it match style of rest of wiki are welcome


fufexan (commented):
The contents of those directories are bound to change.

I created the index files in order to fix categories not being shown.

We can now use the proper fix from https://github.com/imfing/hextra/pull/325.

PR #571, opened 2024-04-02
faq: add method to disable middle click paste

GuillaumeLagrange:
The primary method disables middle click altogether, and the alternative one seems to be very power consuming.

This method only clears the primary buffer when middle clicking, and keeps middle click functionality, instead of doing so everytime the primary buffer is changed. Let me know what you think.

### DIFF
--- a/pages/FAQ/_index.md
+++ b/pages/FAQ/_index.md
@@ -247,13 +247,18 @@ env = XDG_CURRENT_DESKTOP,Hyprland
 
 ### How to disable middle-click paste?
 
-You can simply intercept the middle-click action all together, via hyprland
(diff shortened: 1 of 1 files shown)


fufexan (approved):
LGTM.

@end-4 your thoughts?

end-4:
it works
it's good
image

end-4:
actually after a while using this i find it inconsistent. sometimes the buffer clear isn't fast enough...
but it's still a good method

Donty10:
Is this part of the FAQ needed anymore since there is now a toggle for middle click paste in v0.41.0 which has been documented in the Variables section?

PR #578, opened 2024-04-07
input:off_window_axis_events values

dranull:
hyprwm/Hyprland#4177

Am I allowed to use line-breaks?


vaxerski:
> Am I allowed to use line-breaks?

idfk but I'd rather not

vaxerski (approved):

PR #585, opened 2024-04-09
workspace: add group flag in windowCount selector

sungyoonc:
Wiki MR for hyprwm/Hyprland#5499

- Add `g` flag to `windowCount` prop in workspace selectors.
- Add example for the `windowCount` prop.


vaxerski (approved):
thanks

PR #589, opened 2024-04-10
Popups at cursor position

nnra6864:
Provided a fix for opening popups(such as Unity color picker/material selector) at cursor position and therefore not disappearing when moving the pointer towards them.


vaxerski (commented):
pleae mention to use this only for xwayland (no need for wayland)

nnra6864:
Done.

vaxerski (approved):

PR #593, opened 2024-04-12
Remove ~, replace with full path

DaraghD:
It seems that [wlroots can't](https://github.com/hyprwm/Hyprland/issues/5553) expand the ~ for user home directory, so full path must be used without it.
 pinging @kiecla as they initially wrote it and there may be some misunderstanding, assuming it worked for them when it was wrote.

### DIFF
--- a/pages/Configuring/Multi-GPU.md
+++ b/pages/Configuring/Multi-GPU.md
@@ -65,7 +65,6 @@ If instead you would like to use another GPU, you must first create a symlink to
 the card from the previous section.
 
 It is not possible to use `~/.config/hypr/card` as wlroots will not expand it correctly.  
-
 You must include full path e.g `/home/<user>/.config/hypr/card` 
 ```
 ln -sf /dev/dri/pci-0000:06:00.0-card /home/<user>/.config/hypr/card
--- a/pages/Configuring/Multi-GPU.md
+++ b/pages/Configuring/Multi-GPU.md
@@ -64,8 +64,11 @@ required. wlroots will set `WLR_DRM_DEVICES` to the integrated GPU by default.
 If instead you would like to use another GPU, you must first create a symlink to
 the card from the previous section.
 
+It is not possible to use `~/.config/hypr/card` as wlroots will not expand it correctly.  
+
+You must include full path e.g `/home/<user>/.config/hypr/card` 
 ```
-ln -sf /dev/dri/pci-0000:06:00.0-card ~/.config/hypr/card
+ln -sf /dev/dri/pci-0000:06:00.0-card /home/<user>/.config/hypr/card
 ```
 
 It is not possible to directly use the `/dev/dri/pci-0000:06:00.0-card` path,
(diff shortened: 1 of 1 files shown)


kiecla:
I didn't test that set up. If I recall I copied that down from somewhere else. My bad. I leave that variable unset and wlroots will default to the iGPU in that case. This PR is a necessary change to what I wrote.
I might suggest telling the user to put the symlink somewhere user-agnostic, i.e. outside of the home directory, so that on systems with more than one user (probably not many nowadays, especially on hyprland) this works without having to access another user's home directory.
/usr/local is a candidate, but it might depend on distribution. I don't think that change is all that important really as most people are completely unaffected by it.

Fxzzi:
Should also try `$HOME`, that would probably work, idk tho

MahouShoujoMivutilde:
Apparently `$HOME` does work:

Tested via fully exiting Hyprland, and trying to run it from a tty with

```
env = WLR_DRM_DEVICES,$HOME/.config/hypr/card
```

where card is `/home/witch/.config/hypr/card -> /dev/dri/by-path/pci-0000:09:00.0-card`

\- works fine.

Deliberately providing the wrong path, as expected, makes it fail with an error about failing to create the backend, so I know it actually uses the card from the path instead falling back to whatever works.

---

Also, @DaraghD, it says in one section

> run `ls -l /dev/dri/by-path`

but the section right after, about the symlink, uses

> `ln -sf /dev/dri/pci-0000:06:00.0-card ...`

Notice the missing `by-path`.

For me, `/dev/dri` looks like

```
/dev/dri
├── by-path
│   ├── pci-0000:09:00.0-card -> ../card0
│   └── pci-0000:09:00.0-render -> ../renderD128
├── card0
└── renderD128
```

And [here](https://github.com/hyprwm/Hyprland/issues/5553#issuecomment-2050469991) is `by-path` too.

DaraghD:
> Apparently `$HOME` does work:
> 
> Tested via fully exiting Hyprland, and trying to run it from a tty with
> 
> ```
> env = WLR_DRM_DEVICES,$HOME/.config/hypr/card
> ```
> 
> where card is `/home/witch/.config/hypr/card -> /dev/dri/by-path/pci-0000:09:00.0-card`
> 
> - works fine.
> 
> Deliberately providing the wrong path, as expected, makes it fail with an error about failing to create the backend, so I know it actually uses the card from the path instead falling back to whatever works.
> 
> Also, @DaraghD, it says in one section
> 
> > run `ls -l /dev/dri/by-path`
> 
> but the section right after, about the symlink, uses
> 
> > `ln -sf /dev/dri/pci-0000:06:00.0-card ...`
> 
> Notice the missing `by-path`.
> 
> For me, `/dev/dri` looks like
> 
> ```
> /dev/dri
> ├── by-path
> │   ├── pci-0000:09:00.0-card -> ../card0
> │   └── pci-0000:09:00.0-render -> ../renderD128
> ├── card0
> └── renderD128
> ```
> 
> And [here](https://github.com/hyprwm/Hyprland/issues/5553#issuecomment-2050469991) is `by-path` too.

The documentation says explicitly not to use that, however it may be wrong again. I only told the person in the issue to do that to see if it would work as a hack, I couldn't test the other way so couldn't confirm if it worked or not.
Possibly should also change the documentation from /home/user to $HOME, up to whatever the maintainer prefers-  maybe mention both?

Fxzzi:
> change the documentation from /home/user to $HOME, up to whatever the maintainer prefers- maybe mention both?

I'm not a maintainer however I feel leaving just `$HOME` should suffice. I have not seen one circumstance where it's not set, and also allows for portability - e.g. if someone copies dotfiles of another / username changes the config will still work without manually needing to change.

MahouShoujoMivutilde:
> The documentation says explicitly not to use that, however it may be wrong again.

Well it's wrong in that there is to `pci...` thing there at all, because it's in its own subdirectory.

I'm pretty sure `by-path` was just forgotten, because as far as I can tell `/dev/dri` only supposed to contain `cardN`, `render...`, not pci, those are inside `by-path`, from everywhere I can find.

I agree, if $HOME works - no reason not to use it.

fufexan (approved):

PR #598, opened 2024-04-13
windowrulev2: fix typo

leon-erd:
`initialtitle` (without capital T) actually matches `title` instead (see https://github.com/hyprwm/Hyprland/issues/5550).
Also explicitly say that the supported fields are only for v2 since they do not work for v1 and I at least oversaw they were listed in the v2 section.

### DIFF
--- a/pages/Configuring/Window-Rules.md
+++ b/pages/Configuring/Window-Rules.md
@@ -65,13 +65,13 @@ terminals.
 
 {{< /callout >}}
 
-For now, the supported fields are:
+For now, the supported fields for V2 are:
 
 ```ini
 class - class regex 
 title - title regex
 initialclass - initialClass regex
-initialtitle - initialTitle regex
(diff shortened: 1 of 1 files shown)


fufexan:
Might be a good idea to also capitalize `initialClass`.

leon-erd:
I don't think this is implemented yet (see https://github.com/hyprwm/Hyprland/issues/4965).
For me its hard to test if something matches `class` or `initialclass` or `initialClass` because all my apps have the same `class` and `initialClass` reported by `hyprctl`.
There is also a remark specifically for v2 `initialclass`
https://github.com/hyprwm/hyprland-wiki/blob/d76c2223012370f290f8255104db3640ad7ba33b/pages/Configuring/Window-Rules.md?plain=1#L87

fufexan (approved):

PR #608, opened 2024-04-21
update chromium/electron section

MQ1995:


fufexan (commented):
Might be helpful to include https://github.com/hyprwm/hyprland-wiki/blob/main/pages/Nvidia/_index.md?plain=1#L168, or just removing the whole section from the Nvidia page and adding the info here.

@Fxzzi thoughts?

MQ1995:
> Might be helpful to include https://github.com/hyprwm/hyprland-wiki/blob/main/pages/Nvidia/_index.md?plain=1#L168, or just removing the whole section from the Nvidia page and adding the info here.

I agree, I didn't even know that section contained similar info, I didn't read it since I don't have a Nvidia gpu.
The part about vscodium and obsidian.md would have been useful to me.

Fxzzi:
> @Fxzzi thoughts?

Yep, I was thinking this too. Lets move the electron / cef info over to this PR and replace the nvidia side with a link to this page. That seems like the best course of action.

thejch:
Worth noting that the electron-flags.conf trick is because Arch Linux uses a wrapper for electron and vscode that applies the flags, so this method may not work on other distros

Fxzzi:
> Worth noting that the electron-flags.conf trick is because Arch Linux uses a wrapper for electron and vscode that applies the flags, so this method may not work on other distros

The electron env var will work regardless of distro as long as the app is using a relatively recent electron version (not sure exactly which version added this envvar, might be E27 or E28)

marcorichetta:
> > Worth noting that the electron-flags.conf trick is because Arch Linux uses a wrapper for electron and vscode that applies the flags, so this method may not work on other distros
> 
> The electron env var will work regardless of distro as long as the app is using a relatively recent electron version (not sure exactly which version added this envvar, might be E27 or E28)

Electron >= 28 should work with the env variable per [this](https://wiki.archlinux.org/title/Wayland#Environment_variable).

FYI, I didn't set any flags, just the env variable `ELECTRON_OZONE_PLATFORM_HINT=auto` and it worked (Vscode, Obsidian). 

OS: EndeavourOS

https://www.electronjs.org/docs/latest/api/environment-variables#electron_ozone_platform_hint-linux

PR #613, opened 2024-04-24
Update Master-Layout.md

nekrogenesis:
The default value for `mfact` seems to be `0.55`. [Hyprland/src/ConfigManager.cpp](https://github.com/hyprwm/Hyprland/blob/34413d1f367b56e17614ee0455b6bf0b61a5aed1/src/config/ConfigManager.cpp#L427)


vaxerski (approved):
correct

PR #615, opened 2024-04-24
Suggest to also bind keyboard keys instead of mouse clicks for Touchpads

6543:
address https://github.com/hyprwm/Hyprland/issues/5727


vaxerski (approved):
ok

PR #618, opened 2024-04-26
GDK_BACKEND separator fix and try any backend

anewdi:
GDK_BACKEND only supports comma separator. Also added try for any gdk backend after x11.


vaxerski (approved):
thanks

PR #634, opened 2024-05-05
add arch dependency: hyprwayland-scanner

sambergo:


vaxerski:
it's not yet in the repos, though.

czM1K3:
But literally few lines below it is the line to clone latest git... I's bit confusing.

vaxerski:
I've added it to instructions already

image

PR #639, opened 2024-05-07
Simplify keybind generation code for workspaces

simsam8:
Change workspace keybind generation to use keycodes instead of  numbers. This makes the code snippet work with different layouts. Removed redundant code.

### DIFF
--- a/pages/Nix/Hyprland on Home Manager.md
+++ b/pages/Nix/Hyprland on Home Manager.md
@@ -137,13 +137,10 @@ module, or the flake-based Home Manager module.
         # binds $mod + [shift +] {1..10} to [move to] workspace {1..10}
         builtins.concatLists (builtins.genList (
             x: let
-              ws = let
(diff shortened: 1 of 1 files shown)


pages/Nix/Hyprland on Home Manager.md
```diff
@@ -137,13 +137,10 @@ module, or the flake-based Home Manager module.
         # binds $mod + [shift +] {1..10} to [move to] workspace {1..10}
         builtins.concatLists (builtins.genList (
             x: let
-              ws = let
-                c = (x + 1) / 10;
-              in
-                builtins.toString (x + 1 - (c * 10));
+              ws = builtins.toString (x + 10);
```
fufexan, line 140: Based on the [keycodes](https://github.com/xkbcommon/libxkbcommon/blob/master/include/xkbcommon/xkbcommon-keysyms.h#L645) shouldn't this be `+ 48`?

PR #666, opened 2024-05-29
added context for vertical monitor stacking

bartilg:
Documentation lacked a clear example of placing monitors vertically. I found it rather confusing, as I believed that a larger value would place a monitor higher, similar to mathematical coordinates. I believe this example may clarify this for anyone else having a similar confusion.

### DIFF
--- a/pages/Configuring/Monitors.md
+++ b/pages/Configuring/Monitors.md
@@ -53,6 +53,25 @@ monitor=DP-1, 1920x1080, 0x0, 1
 monitor=DP-2, 1920x1080, -1920x0, 1
 ```
 
+Recall the `position` is calculated from the top-left corner. Thus, a larger number
+places a monitor lower relative to the others.
+
+For example:
+
+```ini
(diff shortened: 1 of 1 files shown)


pages/Configuring/Monitors.md
```diff
@@ -53,6 +53,25 @@ monitor=DP-1, 1920x1080, 0x0, 1
 monitor=DP-2, 1920x1080, -1920x0, 1
 ```
 
+Recall the `position` is calculated from the top-left corner. Thus, a larger number
+places a monitor lower relative to the others.
+
```
fufexan, line 58: The second sentence doesn't really make sense here. Anything with `-` in front will go above, and anything without will go below. It's an inverted Y cartesian coordinate system.

bartilg: I'll rephrase to improve the wording and resubmit. I spent a good bit of time confused about the inverted nature of the y-axis and think an example would be beneficial for the documentation.

fufexan (approved):
LGTM, thanks!

PR #680, opened 2024-06-13
Update installation instructions for Void Linux

Makrennel:
Made some changes to make Hyprland easier to install on Void Linux, and updated the guide to reflect that.


fufexan (approved):
LGTM, thanks!

PR #699, opened 2024-06-25
Dispatchers: fix typo

leon-erd:
@fufexan When submitting https://github.com/hyprwm/hyprland-wiki/pull/696 I thought that referencing would work like this https://github.com/hyprwm/hyprland-wiki/blob/33bd679037ce9bba7a11cd202053f671a4d0414c/pages/Configuring/Dispatchers.md?plain=1#L185 because both files are in the same folder. But as I found out now when I clicked on the link in the wiki it doesn't. 
Is this now correct?


fufexan (approved):
My bad, I didn't catch it yesterday. Yes, it is correct now.

(I should write a contributing guide)

vaxerski:
@fufexan you gon merge?

fufexan:
forgor

PR #712, opened 2024-07-12
Fix incorrect workspace rules in example code

echus:
Replaces "name" rule with "defaultName" in workspace rule examples.


vaxerski:
named workspaces exist sir

PR #731, opened 2024-08-01
Installation: Add Chaotic AUR and `downgrade` mention

Visual-Dawg:
Just sharing what I find useful.

The Chaotic Aur is awesome and downgrade makes it quick and easy to, well, downgrade, especially with the Chaotic Aur.

### DIFF
--- a/pages/Getting Started/Installation.md
+++ b/pages/Getting Started/Installation.md
@@ -19,7 +19,7 @@ might have **major** issues running Hyprland.
 
 ## Installation
 
-Installing Hyprland is very easy. Simply install it with your package manager 
+Installing Hyprland is very easy. Simply install it with your package manager
 (if there is a Hyprland package available) or install/build it yourself.
 
 {{< callout >}}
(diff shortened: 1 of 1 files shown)


alba4k:
> we recommend

chaotic-aur is a really nice thing to have, but I wouldn't officially recommend it as it isn't a great idea in principle. Of course it's not a complete anarchy, but I still think that sentence should be reworded sligthly

Visual-Dawg:
True, I changed it

alba4k (changes requested):
pre-built not pre-build

also, the part after the comma doesn't really make sense grammatically


"and be able to perform easy downgrades", maybe?

alba4k:
It should probably also be made clear that this is all regarding Arch Linux, as NixOS is also officially supported

Visual-Dawg:
This is within the Arch Linux install card, so it should be clear enough

alba4k:
You're right, my bad ahah

pages/Getting Started/Installation.md
```diff
@@ -42,6 +42,10 @@ hyprland-git (AUR) - compiles from latest source
 hyprland - binary x86 tagged release
 ```
 
+If you decide to use the `git` version, you can use the [Chaotic Aur](https://aur.chaotic.cx/) to get pre-built binaries.
```
alba4k, line 45: maybe change this part to

> the `-git` version from the AUR

alba4k (changes requested):

fufexan (approved):
LGTM

vdawg-git:
@alba4k  Good idea, I also experienced this. 
Vax could also just bumb all package versions when there is an ABI break, but I suppose adding a warning or removing the chaotic AUR is better

PR #758, opened 2024-08-23
Mention the use of dconf to fix errors with gtk themes, since gsettings doesn't work and is what the sway devs suggest

Kommynct:
Literally spent hours trying to use gsettings like sway suggests before I found it gsettings just doesn't work properly on nixos and you need to use dconf.

If this was written here it would've saved me so much time.

### DIFF
--- a/pages/Nix/Hyprland on NixOS.md
+++ b/pages/Nix/Hyprland on NixOS.md
@@ -147,7 +147,7 @@ in {
 If your themes for mouse cursors, icons or windows don't load correctly, see the
 relevant section in [Hyprland on Home Manager](../Hyprland-on-Home-Manager).
 
-If you prefer not to use home manager, you can also resolve the issues with GTK themes using dconf like so:
+If you prefer not to use Home Manager, you can also resolve the issues with GTK themes using dconf like so:
 
 exec-once = dconf write /org/gnome/desktop/interface/gtk-theme "'Adwaita'"
 exec-once = dconf write /org/gnome/desktop/interface/icon-theme "'Flat-Remix-Red-Dark'"
--- a/pages/Nix/Hyprland on NixOS.md
+++ b/pages/Nix/Hyprland on NixOS.md
@@ -147,10 +147,13 @@ in {
 If your themes for mouse cursors, icons or windows don't load correctly, see the
 relevant section in [Hyprland on Home Manager](../Hyprland-on-Home-Manager).
 
-If you prefer not to use Home Manager, you can also resolve the issues with GTK themes using dconf like so:
+If you prefer not to use Home Manager, you can also resolve the issues with GTK
+themes using dconf like so:
 
+```ini
 exec-once = dconf write /org/gnome/desktop/interface/gtk-theme "'Adwaita'"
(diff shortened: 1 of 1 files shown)


pages/Nix/Hyprland on NixOS.md
```diff
@@ -146,3 +146,11 @@ in {
 
 If your themes for mouse cursors, icons or windows don't load correctly, see the
 relevant section in [Hyprland on Home Manager](../Hyprland-on-Home-Manager).
+
+If you prefer not to use home manager, you can also resolve the issues with GTK themes using dconf like so:
```
fufexan, line 150:
```suggestion
If you prefer not to use Home Manager, you can also resolve the issues with GTK themes using dconf like so:
```

pages/Nix/Hyprland on NixOS.md
```diff
@@ -146,3 +146,11 @@ in {
...
+
+exec-once = dconf write /org/gnome/desktop/interface/gtk-theme "'Adwaita'"
+exec-once = dconf write /org/gnome/desktop/interface/icon-theme "'Flat-Remix-Red-Dark'"
+exec-once = dconf write /org/gnome/desktop/interface/document-font-name "'Noto Sans Medium 11'"
+exec-once = dconf write /org/gnome/desktop/interface/font-name "'Noto Sans Medium 11'"
+exec-once = dconf write /org/gnome/desktop/interface/monospace-font-name "'Noto Sans Mono Medium 11'"
```
fufexan, line 156:
```suggestion
\`\`\`ini
exec-once = dconf write /org/gnome/desktop/interface/gtk-theme "'Adwaita'"
exec-once = dconf write /org/gnome/desktop/interface/icon-theme "'Flat-Remix-Red-Dark'"
exec-once = dconf write /org/gnome/desktop/interface/document-font-name "'Noto Sans Medium 11'"
exec-once = dconf write /org/gnome/desktop/interface/font-name "'Noto Sans Medium 11'"
exec-once = dconf write /org/gnome/desktop/interface/monospace-font-name "'Noto Sans Mono Medium 11'"
\`\`\`
```

The `\` are there to escape the actual ```

fufexan (commented):
dconf keeps its state through reboots, so adding this to `exec-once` is perhaps redundant, as the commands only need to be run when actually changing the values.

Still, I don't know of a better way to do this except activation scripts, which is another beast. So let's leave it at this.

Kommynct:
Sounds good, thanks!

pages/Nix/Hyprland on NixOS.md
```diff
@@ -146,3 +146,13 @@ in {
 
 If your themes for mouse cursors, icons or windows don't load correctly, see the
 relevant section in [Hyprland on Home Manager](../Hyprland-on-Home-Manager).
+
+If you prefer not to use Home Manager, you can also resolve the issues with GTK themes using dconf like so:
+
+\`\`\`ini
```
fufexan, line 152: I meant without the actual `\`

Kommynct: I pushed resolve I think that automatically adds your changes, sorry if it does not, I'm not very familiar with github.

fufexan (approved):

PR #761, opened 2024-08-24
refactor: use table for window rules v2 field

sungyoonc:
Changed the window rules v2 fields' code block to a table.


fufexan (approved):
LGTM

PR #770, opened 2024-09-07
wiki: add exec-shutdown

Trimutex:
wiki PR for https://github.com/hyprwm/Hyprland/pull/7683


Trimutex:
Might want to double check I'm not missing an entry where this might be important tho

vaxerski (approved):

PR #771, opened 2024-09-08
Fix hyprland -> Hyprland on Master Tutorial

Polarzincomfrio:


fufexan:
Not really necessary, as `hyprland` is a symlink to the `Hyprland` binary.

Polarzincomfrio:
but the `hyprland` and `hyprland -h` command's will never work since the correct one that needs to be ran on the tty is `Hyprland`

fufexan:
Why will it never work? I just tried it and it works.

Polarzincomfrio:
So on your setup, `hyprland` works because of how it’s linked, but on mine, it might not by default. That’s why using `Hyprland` is normally a better practice, since it’s the real command and avoids issues across different setups. 
image

fufexan:
Ok, sure.

fufexan (approved):

PR #781, opened 2024-09-23
chore: add HYPRLAND_CONFIG to the "Environment variables" page

amadalusia:
Reflecting a change that was made in the main branch of Hyprland.


amadalusia:
Personally I am fine with this PR not being merged until 0.43.1 because it seems as if the wiki only tracks per new version and never when a new commit on main happens.

vaxerski (approved):

vaxerski:
> Personally I am fine with this PR not being merged until 0.43.1 because it seems as if the wiki only tracks per new version and never when a new commit on main happens.

quite the opposite

PR #784, opened 2024-09-25
Add install instructions for Ximper Linux

fiersik:
Packages of the Hyprland ecosystem are available in Ximper Linux.

It is planned to create a separate build based on this window manager in the near future.

### DIFF
--- a/pages/Getting Started/Installation.md
+++ b/pages/Getting Started/Installation.md
@@ -296,6 +296,32 @@ apk add hyprland
 
 {{% /details %}}
 
+{{% details title="Ximper*" closed="true" %}}
+
+Install from the Sisyphus:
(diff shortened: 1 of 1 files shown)


vaxerski:
I am unsure if ximper is significant enough to be worth putting into the already cluttered installation combo

X1mper:
> I am unsure if ximper is significant enough to be worth putting into the already cluttered installation combo
Isn't it important for hyprland developers to work with the community?
I think every distribution is important

fufexan (approved):

fufexan:
The installation combo may be cluttered but we can split it into multiple pages.

vaxerski:
fair neough

PR #787, opened 2024-09-28
Add flameshot as a screenshot tool option

tonyd33:
Flameshot is a popular and well-featured tool used in X so it's worth mentioning in the FAQ on screenshot tools.

Although it doesn't work out of the box on Wayland, a few windowrules can be added to fix the incompatibility which have been added to the FAQ.

### DIFF
--- a/pages/FAQ/_index.md
+++ b/pages/FAQ/_index.md
@@ -93,6 +93,10 @@ windowrulev2 = pin, class:^(flameshot)$
 # set this to your leftmost monitor id, otherwise you have to move your cursor to the leftmost monitor
 # before executing flameshot
 windowrulev2 = monitor 1, class:^(flameshot)$
+
+# ctrl-c to copy from the flameshot gui gives warped images sometimes.
+# piping it into wl-copy seems to always work though
+bind = ..., exec, flameshot gui --raw | wl-copy
 ```
 
 **Option 3:** You can also use hyprshot, more info [here](https://github.com/Gustash/Hyprshot).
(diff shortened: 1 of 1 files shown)


vaxerski:
idk, it's very broken atm and I wouldn't recommend it

tonyd33:
> idk, it's very broken atm and I wouldn't recommend it

How so? It's working pretty much fine for me on v12.1.0 (14a13677) from aur/flameshot-git with the settings I gave.

The exception I just realized was that copying to clipboard can give warped results, but it seems to be fixed by calling `flameshot gui --raw | wl-copy` instead of just `flameshot gui`.

tonyd33:
I guess you can look at it that way.

I think it's a shame to not include because the other options require much more tinkering to get features on-par with flameshot, which just requires a little configuration that'd be documented in the FAQ to work.

We could move flameshot to the last option and add emphasize the disclaimer that flameshot isn't working as intended on Wayland but the configuration fixes it or we could close this PR until flameshot implements the fixes themselves if we still think it isn't worth using. Your call

vaxerski:
> We could move flameshot to the last option and add emphasize the disclaimer that flameshot isn't working as intended on Wayland but the configuration fixes it

that's fine. Also worth noting that on some setups (hidpi / multimon  etc) I've heard very mixed results with flameshot

tonyd33:
> Also worth noting that on some setups (hidpi / multimon etc) I've heard very mixed results with flameshot

Yeah, I'm on a HiDPI and multi-monitor setup, and I think I've come across why they have mixed results. The configuration is what solved it for me though.

I moved the Flameshot option down and added more emphasis that it's broken and users should use it at their own discretion.

vaxerski (approved):

PR #790, opened 2024-09-29
Specified bad XDG env could break desktop portal

BlackFuffey:
I've recently encountered a seemingly not very well documented issue of bad XDG env breaking desktop portal. I specified said issue in wiki so others can have an easier time dealing with this issue


vaxerski (approved):
sure, but in the case of XDG_CURRENT_DESKTOP, I've added a check in -git anyways that will pop up a notification if its wrong

PR #800, opened 2024-10-10
Update Environment-variables.md

p3nj:
Remove Duplicate Lines under [XDG Specifications]


fufexan (commented):
Looks good, but we could also mention that Hyprland pre-checks their existence, and if not found, sets them to the listed values.

PR #814, opened 2024-10-19
variables: add the `general:snap` subcategory

myQwil:
Wiki entry for https://github.com/hyprwm/Hyprland/pull/8088

Moved the callout section regarding subcategories, as this is now the first subcategory that users will see when scrolling down the page.


vaxerski:
it is still invalid

vaxerski (approved):

PR #818, opened 2024-10-21
hyprlock: Clarify that XDG env can affect the fade out behavior.

aaron-randreth:
Having a badly set XDG env variable is currently quite confusing as there is no indication this affects the way hyprlock fades out (but not in).

See hyprwm/hyprlock#521


vaxerski (commented):
this should be worded differently. Setting it to `Hyprland` forcefully on other environments might produce unwanted effects.

I'd just go with "if you are not on hyprland, or your XDG_CURRENT_DESKTOP is not Hyprland..."

vaxerski (approved):

PR #828, opened 2024-10-25
Fix 404 in master tutorial

Damianu:
Link previously led to https://wiki.hyprland.org/Configuring/Configuring-Hyprland which doesn't exist. 
I think it was meant to point to https://wiki.hyprland.org/Configuring instead.


fufexan (approved):
Forgot to update the link, thanks.

PR #836, opened 2024-11-01
FIX C++ std version on wiki

ismagilli:
C++26 haven't been released and, therefore, doesn't supported by any compiler =)


fufexan:
Please refrain from claiming such things without being aware of the situation.

vaxerski:
IMG_20200717_234542

PR #840, opened 2024-11-04
Add hyprmacros to list of useful utilities

lunatic-gh:
Just a "small" python program i wrote for hyprland, allowing creating per-app macros in a simple way.

configuring macros is intended to be similar to hyprland, eg `bind = LCTRL LSHIFT, H, send, A:DOWN A:UP DELAY:500 B:DOWN B:UP`, while config `hyprmacros.conf` lies in the same folder as hyprland's default config.


lunatic-gh:
I just saw that there are a few issues in it that might take some time to fix.

I'm going to close this for now, and might re-open someday.

PR #842, opened 2024-11-04
Add mention about possibility to start hypridle, hyprpaper and hyprpolkit with system services.

izmyname:
Wording may not be the best, but I think it's worth to mention systemd services, as an alternative to exec-once=. Anyway, suggestions are welcome.

### DIFF
--- a/pages/Hypr Ecosystem/hypridle.md
+++ b/pages/Hypr Ecosystem/hypridle.md
@@ -9,7 +9,7 @@ hypridle is Hyprland's idle management daemon.
 
 Configuration is done via the config file at `~/.config/hypr/hypridle.conf`. A
 config file is required; hypridle won't run without one. To run hypridle at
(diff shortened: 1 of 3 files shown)


fufexan (changes requested):
> If you start Hyprland with a systemd service

I don't think this is necessary. I don't start hyprland as a systemd service but still use user services for hypr* apps.

izmyname:
From my experience, it's only possible to start systemd services from hyprland.conf, in your case as Hyprland, itself, doesn't provide `graphical-session.target`, required for them to be enabled on startup.

This was the reason why I started to work on hyprland-session.service integration, in the first place.

fufexan:
Ok, makes sense then. Still, I think the wording should be "if you start Hyprland **as** a systemd service".

fufexan (approved):
Generally LGTM.

PR #850, opened 2024-11-07
Add uwsm instructions

izmyname:
https://github.com/hyprwm/Hyprland/pull/8376 and https://github.com/hyprwm/Hyprland/pull/8339 discussion

Probably, need to add another page to mention uwsm with short instructions. Startup or something

Corrections, reviews and suggestions are welcome.

### DIFF
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -85,7 +85,7 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
-uwsm users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
+[uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
 
 It's also strongly adviced to replace `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
 
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -91,6 +91,17 @@ It's also strongly adviced to replace `exit` dispatcher with `loginctl terminate
 
 {{< /callout >}}
 
+{{< callout type=warning >}}
+
+It is NOT recommended to set DPMS with a keybind directly, as it might cause
+undefined behavior. Instead, consider something like
+
+```ini
+bind = MOD, KEY, exec, sleep 1 && hyprctl dispatch dpms off
+```
+
+{{< /callout >}}
+
 ### Grouped (tabbed) windows
 
 Hyprland allows you to make a group from the current active window with the
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -87,7 +87,7 @@ layout pages (See the sidebar).
 
 [uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
 
-It's also strongly advised to replace the `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
+It's also strongly advised to replace the `exit` dispatcher with `exec, loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
 
 {{< /callout >}}
 
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -85,7 +85,7 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
-[uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
+[uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, or terminating Hyprland process directly, as exiting Hyprland this way removes it from under its clients and may cause some systemd services to enter inconsistent states causing problems in the next session. Use `exec, loginctl terminate-user ""` instead.
 
 It's also strongly advised to replace the `exit` dispatcher with `exec, loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
 
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -85,12 +85,9 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
-It is NOT recommended to set DPMS with a keybind directly, as it might cause
-undefined behavior. Instead, consider something like
+uwsm users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
 
-```ini
-bind = MOD, KEY, exec, sleep 1 && hyprctl dispatch dpms off
-```
+It's also strongly adviced to replace `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
 
 {{< /callout >}}
 
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -87,7 +87,7 @@ layout pages (See the sidebar).
 
 [uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
 
-It's also strongly adviced to replace `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
+It's also strongly advised to replace the `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
 
 {{< /callout >}}
 
--- a/pages/Getting Started/Master-Tutorial.md
+++ b/pages/Getting Started/Master-Tutorial.md
@@ -47,7 +47,7 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instrustions.
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
 
 
 {{< callout type=warning >}}
--- a/pages/Getting Started/Master-Tutorial.md
+++ b/pages/Getting Started/Master-Tutorial.md
@@ -47,7 +47,7 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as: [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support, launching any application as a systemd-unit with `uwsm app` helper, and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support, launching any application as a systemd-unit with `uwsm app` helper, and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
 
 
 {{< callout type=warning >}}
--- a/pages/Getting Started/Master-Tutorial.md
+++ b/pages/Getting Started/Master-Tutorial.md
@@ -47,27 +47,7 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm), either by adding the following code in shell profile.
-
-```
-if uwsm check may-start && uwsm select; then
-	exec systemd-cat -t uwsm_start uwsm start default
-fi
-```
-
-
-or by choosing `hyprland (uwsm-managed)` entry in a display manager menu.  This method provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar).
-
-{{< callout type=info >}}
-uwsm can launch Hyprland immediately, bypassing the compositor selection menu. To have Hyprland start automatically after logging in, add the following code to your shell profile.
-
-
-```
-if uwsm check may-start; then
-    exec uwsm start hyprland.desktop
-fi 
-```
-{{< /callout >}}
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended way to launch Hyprland on systemd-based distros. See [uwsm](../../Systemd-start) page for further instrustions.
 
 
 {{< callout type=warning >}}
--- a/pages/Getting Started/Master-Tutorial.md
+++ b/pages/Getting Started/Master-Tutorial.md
@@ -47,7 +47,7 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm), either by adding the following command in shell profile.
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm), either by adding the following code in shell profile.
 
 ```
 if uwsm check may-start && uwsm select; then
--- a/pages/Getting Started/Master-Tutorial.md
+++ b/pages/Getting Started/Master-Tutorial.md
@@ -47,7 +47,7 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended way to launch Hyprland on systemd-based distros. See [uwsm](../../Systemd-start) page for further instrustions.
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instrustions.
 
 
 {{< callout type=warning >}}
--- a/pages/Getting Started/Master-Tutorial.md
+++ b/pages/Getting Started/Master-Tutorial.md
@@ -47,7 +47,7 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support, launching any application as a systemd-unit with `uwsm app` helper and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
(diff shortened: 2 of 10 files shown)


izmyname:
Also, @Vladimir-csp tagging you to review in case I forgot to mention something.

fufexan (commented):
I don't even know why these two are here. They're automatically set by Hyprland. As for `XDG_SESSION_DESKTOP`, what uses it?

https://github.com/hyprwm/hyprland-wiki/blob/ac31e2dd4f66c322e1a492d1bd6db3a5e965f92d/pages/Configuring/Environment-variables.md?plain=1#L64-L65

Vladimir-csp:
I think [this section](https://github.com/Vladimir-csp/uwsm#3-applications-and-slices) needs to be integrated in some shape or form. The "full experience" is to place compositor unit in `session.slice` and run apps as units via `uwsm app` or [analogs](https://github.com/Vladimir-csp/app2unit), this way `uuctl` and `fumon` tools also become more useful.

pages/Getting Started/Master-Tutorial.md
```diff
@@ -47,7 +47,15 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, by typing `systemctl --user start hyprland-session.service` in tty. This method provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar).
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm), either by adding the following command in shell profile.
```
Vladimir-csp, line 50: "code" or "construct" instead of "command".

pages/Getting Started/Master-Tutorial.md
```diff
@@ -47,7 +47,15 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
...
+
+```
+if uwsm check may-start && uwsm select; then
```
Vladimir-csp, line 53: This example gives selection (whiptail) and ability to opt out to shell. It is perfectly fine, but there are shorter variants that would launch just Hyprland. I.e.:
```
if uwsm check may-start; then
    exec uwsm start hyprland.desktop
fi
```

izmyname: Didn't know about that. I'm going to test it with my shell

Vladimir-csp:
> I don't even know why these two are here. They're automatically set by Hyprland. 

IIRC Hyprland passes them through if not mentioned in config, but already present in the environment (typical for DM/SM setup).

> As for `XDG_SESSION_DESKTOP`, what uses it?

Systemd uses it, described in `man pam_systemd`.

izmyname:
> I think [this section](https://github.com/Vladimir-csp/uwsm#3-applications-and-slices) needs to be integrated in some shape or form. The "full experience" is to place compositor unit in `session.slice` and run apps as units via `uwsm app` or [analogs](https://github.com/Vladimir-csp/app2unit), this way `uuctl` and `fumon` tools also become more useful.

Makes sense to put it on a specific page a mentioned in the initial comment. Although, honestly, I didn't try uwsm app and not 100% sure how it works.

pages/Getting Started/Master-Tutorial.md
fufexan: Is this present in uwsm's readme? I was looking for a way to use uwsm with greetd autologin.

izmyname: https://github.com/Vladimir-csp/uwsm/blob/master/README.md#where-to-launch-from

izmyname:
Oops, I messed up this one.

pages/Getting Started/Master-Tutorial.md
Vladimir-csp: >  I was looking for a way to use uwsm with greetd autologin.

For autologin you can probably use `uwsm start ...` command directly.

izmyname:
IMHO, needs to be rephrased and simplified, so we could add a quick instruction ( see https://wiki.hyprland.org/Useful-Utilities/). Long version is better to be put somewhere in readme on uwsm side, so we could provide a link to it.

Speaking of adding another page for uwsm @fufexan should it be done by me, in this pr, or somehow else> Alternatvely, though, we could put uwsm in "Other" or "Must have" section

fufexan:
You can create a new page under Useful Utilities in this PR.

izmyname:
Oh wait, I just noticed Hyprland wiki already mention s uwsm. https://wiki.hyprland.org/Useful-Utilities/Other/

Should I keep the new page or that mention should be enough?

fufexan:
Keep the new page, and remove the mention from Other.

izmyname:
Okay, done. I'll revisit the page later to add installation instructions and links to other wiki pages, if needed.

pages/Configuring/Dispatchers.md
```diff
@@ -85,6 +85,14 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
+uwsm users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
+
+It's also strongly adviced to replace `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
```
Vladimir-csp, line 90: could use a link to https://github.com/Vladimir-csp/uwsm#how-to-stop here

pages/Configuring/Dispatchers.md
```diff
@@ -85,6 +85,14 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
+uwsm users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
```
Vladimir-csp, line 88: It shouldn't cause those kind of problems, just a bunch of errors in apps due to sudden compositor disappearance. Could you file a bug to uwsm with logs?

pages/Configuring/Dispatchers.md
izmyname: I think it's better to just leave the warning, in this

pages/Configuring/Dispatchers.md
izmyname: I can check it, later, but since compositor-site exit isn't a desirable way to exit the session with uwsm, I think we should leave the warning

Vladimir-csp: Warning/note on how to properly exit is valid.
Scope of consequences needs checking.

izmyname: What logs do you need?

Vladimir-csp: `journalctl --user` when exiting via dispatcher and then starting again.

izmyname: https://github.com/Vladimir-csp/uwsm/issues/51

fufexan (commented):
I think references to uwsm should contain a link to the dedicated page.

Info about starting hyprland via uwsm should also be moved to the dedicated page, and only a link should be given in the master tutorial.

pages/Configuring/Dispatchers.md
```diff
@@ -85,6 +85,14 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
+[uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
+
+It's also strongly adviced to replace `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
```
fufexan, line 90:
```suggestion
It's also strongly advised to replace the `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
```

Also, does `loginctl terminate-user ""` terminate the current user or all users?

pages/Getting Started/Master-Tutorial.md
```diff
@@ -47,7 +47,8 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, by typing `systemctl --user start hyprland-session.service` in tty. This method provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar).
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instrustions.
```
fufexan, line 50:
```suggestion
Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
```

pages/Useful Utilities/Systemd-start.md
```diff
@@ -0,0 +1,67 @@
...
+
+## UWSM
+
+- [Universal Wayland Session Manager](https://github.com/Vladimir-csp/uwsm) is a recommended way to start Hyprland session on systemd distros. uwsm wraps standalone Wayland compositors into a set of Systemd units and provides robust session management including environment, XDG autostart support, bi-directional binding with login session, and clean shutdown.
+
+## Installation
+
+{{% details title="Arch" closed="true" %}}
+
+uwsm is available in AUR. You can install it either with aur helper, or build PKGBUILD manually. Here's an example for yay.
```
fufexan, line 14:
```suggestion
uwsm is available in AUR. You can install it either with an AUR helper, or build PKGBUILD manually. Here's an example for yay.
```

pages/Useful Utilities/Systemd-start.md
```diff
@@ -0,0 +1,67 @@
...
+```sh
+yay -S uwsm
+```
+
+{{% /details %}}
+
+{{% details title="Nix/NixOS" closed="true" %}}
+
+Available [here](https://github.com/NixOS/nixpkgs/tree/master/pkgs/by-name/uw/uwsm/package.nix)
+
```
fufexan, line 25:
```suggestion
\```nix
programs.uwsm = {
  enable = true;
  waylandCompositors.hyprland = {
    binPath = "/run/current-system/sw/bin/Hyprland";
    comment = "Hyprland session managed by uwsm";
    prettyName = "Hyprland";
  };
};
\```

The above option generates a new desktop entry, `hyprland-uwsm.desktop`, which will be available in Desktop Managers.

For more info, read the [option](https://search.nixos.org/options?channel=unstable&show=programs.uwsm.enable&from=0&size=50&sort=relevance&type=packages&query=uwsm).
```

Remove the backslashes, that was the only way to escpe the backticks.

I think I should create a dedicated page under Nix/ instead, as the setup is a bit different in some cases.

pages/Useful Utilities/Systemd-start.md
```diff
@@ -0,0 +1,67 @@
...
+
+{{% /details %}}
+
+
+ {{< callout type=info >}}
```
fufexan, line 30:
```suggestion
{{< callout type=info >}}
```
Same for the callout lines below.

pages/Useful Utilities/Systemd-start.md
```diff
@@ -0,0 +1,67 @@
...
+if uwsm check may-start; then
+    exec uwsm start hyprland.desktop
+fi 
+```
+
+
+If you use a display manager, choose `hyprland (uwsm-managed)` entry in a display manager selection menu.
+
+
+
```
fufexan, line 61: Too many empty lines.

pages/Getting Started/Master-Tutorial.md
```diff
@@ -47,7 +47,8 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, by typing `systemctl --user start hyprland-session.service` in tty. This method provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar).
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
```
Vladimir-csp, line 50: ...plus manage session all-apps-as-units way.

pages/Useful Utilities/Systemd-start.md
```diff
@@ -0,0 +1,73 @@
...
+If you want to bypass compositor selection menu and launch Hyprland directly, use this code in your shell profile, instead.
+
+```
+if uwsm check may-start; then
+    exec uwsm start hyprland.desktop
+fi 
+```
+
+If you use a display manager, choose `hyprland (uwsm-managed)` entry in a display manager selection menu.
+
```
Vladimir-csp, line 86:
```suggestion
## Launching applications inside session

The concept of a session managed by Systemd implies also running applications as units. Uwsm [provides](https://github.com/Vladimir-csp/uwsm#3-applications-and-slices) a helper to do it.

Prefix application startup commands with `uwsm app --`. It also supports launching Desktop Entries by IDs or paths.

Running applications as child processes inside compositor's unit is discouraged.
```

pages/Getting Started/Master-Tutorial.md
fufexan: Regarding that, I see a lot of `\x2d` escapes in the generated units. Is this the fault of NixOS in some way?
image

Vladimir-csp: Believe it or not, this is how it should be. `-` is a special symbol in unit IDs, it should be escaped when not intended for globs. See `man systemd-escape`

fufexan: Do programs have to specify `WantedBy = wayland-session@hyprland\x2duwsm.desktop.target` then? Or does the unescaped name also work?

Although the xdg-autostart route may be the idiomatic way.

pages/Useful Utilities/Systemd-start.md
```diff
@@ -0,0 +1,96 @@
...
+### Using a display manager
+
+If you use a display manager, choose `hyprland (uwsm-managed)` entry in a display manager selection menu.
+
+## Launching applications inside session
+
+The concept of a session managed by Systemd implies also running applications as units. Uwsm [provides](https://github.com/Vladimir-csp/uwsm#3-applications-and-slices) a helper to do it.
+
+Prefix application startup commands with `uwsm app --`. It also supports launching Desktop Entries by IDs or paths.
```
Vladimir-csp, line 93:
```suggestion
Prefix application startup commands with `uwsm app --`. It also supports launching Desktop Entries by IDs or paths. See `man uwsm` or `uwsm app --help`.
```

`--help|-h` stuff is generally useful.

pages/Getting Started/Master-Tutorial.md
Vladimir-csp: `find /run/user/${UID}/systemd/ -name '*\x*'`. Yes, this is canonical representation, also in filenames.

pages/Getting Started/Master-Tutorial.md
```diff
@@ -47,7 +47,8 @@ Please bear in mind 3D acceleration in VMs may be pretty slow.
 
 Hyprland can be executed by typing `Hyprland` in your tty.
 
-Systemd users can also start Hyprland, by typing `systemctl --user start hyprland-session.service` in tty. This method provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support and ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar).
+Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support, launching any application as a systemd-unit with `uwsm app` helper and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
```
Vladimir-csp, line 50:
```suggestion
Systemd users can also start Hyprland, using [uwsm](https://github.com/Vladimir-csp/uwsm). This is the recommended method of launching Hyprland on systemd-based distros, as it provides additional features, such as: [xdg-autostart](https://www.freedesktop.org/software/systemd/man/latest/systemd-xdg-autostart-generator.html) support, launching any application as a systemd-unit with `uwsm app` helper, and the ability to enable services for programs that rely on a graphical session and provide such services (e.g waybar). See [uwsm](../../Systemd-start) page for further instructions.
```
...if I recall English comma rules correctly.

izmyname:
I'm not sure whether my comments were seen, as they were in pending state and I'm not that familiar with github.

@Vladimir-csp I tried to launch `uwsm app  wl-paste --watch cliphist store`, but it resulted in error as uwsm coudn't handle  `--watch cliphist store`. Am I missing sonething? Also, should we use `uwsm app` with flatpak apps, tking into account xdph doesn't implement `org.freedesktop.Background`? And can we one exec-once entry to launch several programs with `uwsm app`, like `exec-once = uwsm app foo & uwsm app bar&`?

fufexan:
@izmyname use `uwsm app -- wl-paste --watch cliphist store`, as mentioned in the man page:

> Always use "--" to disambiguate dashed arguments intended for application itself.

izmyname:
> use `uwsm app -- wl-paste --watch cliphist store`

Uh, I'm an idiot. My bad.  


> Why not have them on separate lines?

Force of habit

Vladimir-csp:
> Also, should we use uwsm app with flatpak apps, taking into account xdph doesn't implement org.freedesktop.Background?

No ATM, since flatpak handles units [its own way](https://github.com/flatpak/flatpak/issues/5870).

> And can we use one exec-once entry to launch several programs with uwsm app, like exec-once = uwsm app foo & uwsm app bar&?

It's shell, so why not.

izmyname:
> No ATM, since flatpak handles units [its own way](https://github.com/flatpak/flatpak/issues/5870).

Plus, things like telegram and easy effects are meant to be autostarted with xdg-autostart, using desktop entries, but xdph doesn't allow that, for now.

Also, another problem I should probably report to uwsm github (will do it tomorrow) -  Hyprland no longer exits via loginctl command, immediately.  All systemd services are stopped, but Hyprland, itself (and flatpak apps) stays active with its default wallpaper placeholder, for like a minute or two, before logging me out. I'll try to reproduce it tomorrow, though.

As for the PR, itself - I'll add an example `exec-once` command in the wiki and, if everything else is ok, it's ready for merging.

pages/Useful Utilities/Systemd-start.md
```diff
@@ -0,0 +1,101 @@
...
+
+The concept of a session managed by Systemd implies also running applications as units. Uwsm [provides](https://github.com/Vladimir-csp/uwsm#3-applications-and-slices) a helper to do it.
+
+Prefix application startup commands with `uwsm app --`. It also supports launching Desktop Entries by IDs or paths. See `man uwsm` or `uwsm app --help`.
+
+An autostart entry example:
+
+```
+exec-once = uwsm app -- mycommand --arg1 --arg2
+```
```
Vladimir-csp, line 99:
````suggestion
Examples for autostart and bind entries:

```
exec-once = uwsm app -- mycommand --arg1 --arg2
bind = SUPER, E, exec, uwsm app -- pcmanfm-qt.desktop
```
````

**Review by Vladimir-csp — APPROVED** [review]
LGTM. Nothing that can't be fixed later.

**Comment by Vladimir-csp** [comment]
> Plus, things like telegram and easy effects are meant to be autostarted with xdg-autostart, using desktop entries, but xdph doesn't allow that, for now.

Why?

**Comment by izmyname** [comment]
> > Plus, things like telegram and easy effects are meant to be autostarted with xdg-autostart, using desktop entries, but xdph doesn't allow that, for now.
> 
> Why?

Because of lacking `org.freedesktop.Background` implemetation. The settings within flatpaks simply won't allow creating a desktop entry.

**Comment by Vladimir-csp** [comment]
Ah, you mean flatpak itself can not put entries into ~/.config/autostart due to portal issues?

**Comment by izmyname** [comment]
Exactly. Additionally, in case of Easy Effects - if I try to access its settings and get a warning about lacking access to Background implementation -  it won't stay as a background process, after I open it, then close, so I have to reset its settings in order to be able both daemonize and access it.

**Comment by izmyname** [comment]
@fufexan tagging for the final (probably) review

File: pages/Configuring/Dispatchers.md [hunk 16]
```diff
@@ -85,6 +85,14 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
+[uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
+
+It's also strongly advised to replace the `exit` dispatcher with `loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
```
fufexan, line 90: I still don't like this. Why not `loginctl terminate-user $(whoami)`?

pages/Hypr Ecosystem/hyprpolkitagent.md
```diff
@@ -14,7 +14,7 @@ or use a different agent, e.g. [KDE's one](https://github.com/KDE/polkit-kde-age
 Add `exec-once = systemctl --user start hyprpolkitagent` to your Hyprland config and restart hyprland.
 (obviously change that to whatever you are using if you are not using the hypr one)
 
-If Hyprland is started as a systemd service, you can autostart for polkit with the command `systemctl --user enable --now hyprpolkitagent.service`.
+If Hyprland is started with [uwsm](../../Systemd-start), you can autostart for polkit with the command `systemctl --user enable --now hyprpolkitagent.service`.
```
fufexan, line 17:
```suggestion
If Hyprland is started with [uwsm](../../Systemd-start), you can autostart the polkit agent with the command `systemctl --user enable --now hyprpolkitagent.service`.
```

pages/Configuring/Dispatchers.md
izmyname: This is the command, listed on uwsm page and what worked for me.

izmyname: And, judging by @Vladimir-csp comment - `loginctl terminate-user $(whoami)` = `loginctl terminate-user ""` in this case.

Vladimir-csp, line 90: Replace dispatcher with actual dispatcher
```suggestion
It's also strongly advised to replace the `exit` dispatcher with `exec, loginctl terminate-user ""` inside `hyprland.conf` keybinds section.
```

pages/Configuring/Dispatchers.md
```diff
@@ -85,6 +85,14 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
+[uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, as exiting Hyprland this way causes various systemd-related problems after restarting the compositor, such xdg-desktop-portal-hyprland and other services do not restart. Use `loginctl terminate-user ""` instead.
```
Vladimir-csp, line 88:
```suggestion
[uwsm](../../Systemd-start) users should avoid using `exit` dispatcher, or terminating Hyprland process directly, as exiting Hyprland this way removes it from under its clients and may cause some systemd services to enter inconsistent states causing problems in the next session. Use `exec, loginctl terminate-user ""` instead.
```

pages/Configuring/Dispatchers.md
fufexan: Yes, this clears my concern.

fufexan (approved):
Formatting looks ok to me. Content-wise it's also good. Will defer to @Vladimir-csp for any content inconsistencies.

izmyname:
I removed `gdb` make dependency for Arch from installation page. Hyprland builds just fine without it, and Arch [dropped](https://gitlab.archlinux.org/archlinux/packaging/packages/hyprland/-/merge_requests/6) it for their PKGBUILD in extra repo.

izmyname:
> Exactly. Additionally, in case of Easy Effects - if I try to access its settings and get a warning about lacking access to Background implementation - it won't stay as a background process, after I open it, then close, so I have to reset its settings in order to be able both daemonize and access it.


Hm. I've just realizing I'm exploiting a vulnerability to bypass portal background permission and daemonize Easy Effects. Curious.

PR #853, opened 2024-11-08
update description for diff groupbar config values to better reflect their role as background colors and not borders.

noornee:
i stumbled on one of `The Linux Cast` video about [tabs in i3](https://youtu.be/sHWYr63mxpg). decided to check if hyprland has similar feature. went through the wiki and found it!

i noticed that the descriptions for the group bar config values were not entirely accurate cos these values actually modify the background color and not the border. so i thought, this was my chance to contribute to Hyprland!


fufexan (approved):
LGTM, thanks!

PR #855, opened 2024-11-09
fix(plugins): missing regex argument for nix search

louis-thevenet:
`nix search` needs at least one regex to match against: `nix search [option...] installable regex...`


fufexan (approved):
This must be new, I didn't even notice.

PR #856, opened 2024-11-09
fix edge case in smart gaps workspace-rule

horriblename:
The previous example has an unhandled edge case: when you have exactly one tiled window and one tiled group, it still triggers the no-gaps rule, which I don't think anyone wants

pre-changes:

image

post-changes:

image


vaxerski:
I'll mirror it to hyprland

PR #858, opened 2024-11-11
Fixed a typo I made in FAQ

nnra6864:
I accidentally wrote `Hprwinwrap` instead of `Hyprwinwrap` and managed to completely miss it.
Not sure if another pr was needed, let me know if I should go about it some other way.


fufexan (approved):
LGTM, I don't think there are any links that need changing.

PR #860, opened 2024-11-11
Fix links

izmyname:
I messed up with the links to other wiki pages, this is an attempt to fix them.


fufexan:
Already fixed. Only the title addition remains.

izmyname:
Okay, I'll make another PR. Didn't see the previous one

PR #863, opened 2024-11-11
Fix  the remaining links to UWSM

izmyname:
These should be the last ones I missed


fufexan:
I really should add a link checker in CI.

PR #868, opened 2024-11-14
Clarify exit command variants

Vladimir-csp:
Resort to `terminate-user` only in case of problems with less blanket options.

### DIFF
--- a/pages/Configuring/Dispatchers.md
+++ b/pages/Configuring/Dispatchers.md
@@ -85,7 +85,7 @@ layout pages (See the sidebar).
 
 {{< callout type=warning >}}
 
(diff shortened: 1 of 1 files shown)


Ghosthree3:
Typo, `boud`.

How does `uwsm stop` compare with `loginctl terminate-session "$XDG_SESSION_ID"`? Will it kill _all_ active graphical session units (ie. if I issue it in tty2 it will kill the tty1 Hyprland instance as well)?
The description on the project page is unclear as to the scope of its effect, but given that the followup option wildcard stops all `wayland-wm@*` services and says it is effectively the same as `uwsm stop` I'm assuming it does.

Vladimir-csp:
Systemd-managed system assumes there is only one graphical session per user. UWSM operates in this paradigm, you will not be able to launch more than one `wayland-wm@.service`.

`uwsm stop`, `systemctl --user stop wayland-wm@.service`, `systemctl --user start wayland-session-shutdown.target` are qeuivalent.

`loginctl terminate-session "$XDG_SESSION_ID"` produces the same result if login session is bound to graphical session.

Ghosthree3:
That makes sense, thank you. I suppose additional instances will just have to be started the traditional way, without uwsm.

fufexan (approved):

PR #874, opened 2024-11-16
Environment-variables - add information - how to export variables, using UWSM

izmyname:
@Vladimir-csp @fufexan

### DIFF
--- a/pages/Configuring/Environment-variables.md
+++ b/pages/Configuring/Environment-variables.md
@@ -3,6 +3,18 @@ weight: 15
 title: Environment variables
 ---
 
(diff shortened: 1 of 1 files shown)


fufexan (approved):

Ghosthree3:
You may want to change https://github.com/hyprwm/hyprland-wiki/blob/main/pages/Useful%20Utilities/Systemd-start.md?plain=1#L61, which currently scrolls right past this information. If I didn't know better I'd assume the info box under `XDG Specifications` was the only information relevant to uwsm on this page.

PR #878, opened 2024-11-18
faq: hyprlock as login screen

sand4rt:
related to; https://github.com/hyprwm/hyprlock/issues/564#issue-2666594217

### DIFF
--- a/pages/FAQ/_index.md
+++ b/pages/FAQ/_index.md
@@ -451,5 +451,5 @@ User=MY_USERNAME
 Add this to `~/.config/hypr/hyprland.conf`:
 
 ```ini
-exec-once=hyprlock || hyprctl dispatch exit
+exec-once=hyprlock --immediate-render || hyprctl dispatch exit
 ```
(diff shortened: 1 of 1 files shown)


pages/FAQ/_index.md
```diff
@@ -435,3 +435,21 @@ This is a side effect of the [decoration:blur:new_optimizations](../Configuring/
 You have two options to resolve it.
 1. Set `decoration:blur:new_optimizations` to `false` - This will preserve the exact same appearance, but may have a slight performance cost.
 2. Set `decoration:blur:ignore_opacity` to `false` - This will drastically affect the appearance, but should maintain the original performance.
+
+### How can I use Hyprlock as a login screen I start my computer?
```
fufexan, line 439:
```suggestion
### How can I use Hyprlock as a login screen when I start my computer?
```

fufexan (commented):
I think it's also worth mentioning that autologin will not automatically unlock keyrings the way display managers usually do.

PaideiaDilemma:
Also Forward from the hyprlock issue:

> Not sure how secure this is BTW?

I can think of two security relevant problems:
- Autologin is a risk in and of itself.
- There is a timeframe where hyprlock has not locked the session yet and Hyprland accepts  normal keybinds and input. That is obviously bad, as you could theoretically kill hyprlock before it actually locks.

PR #887, opened 2024-11-26
binds: add type and default to allow_pin_fullscreen

nnyyxxxx:
image
i think this is correct, i used the information from here
https://github.com/hyprwm/Hyprland/pull/8526


nnyyxxxx:
image
am i tripping or something, why is this true but in configmanager its false
image

vaxerski (approved):

PR #901, opened 2024-12-14
preconfigured setups: remove end-4/dots-hyprland

end-4:
As the author of end-4/dots-hyprland, I feel I cannot maintain it well for now and would like to have it removed from the wiki, at least temporarily, to not cause new users any issues.

### DIFF
--- a/pages/Getting Started/Preconfigured-setups.md
+++ b/pages/Getting Started/Preconfigured-setups.md
@@ -20,14 +20,6 @@ It includes simple gui settings apps, pre-configured feature-rich panels, a welc
 
 ![Image of ML4W dotfiles](https://i.ibb.co/6ydHNt9/screenshot-29-1.png)
 
-## end_4
-
-Like material styling? Want a lot of great apps? Don't mind a tiny bit of tinkering? end-4 has you covered.
-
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
thank you for your work in the past!

clsty:
@vaxerski 2 months has passed. After some PRs merged, commits pushed and issues solved in end-4/dots-hyprland (especially [using virtual environment for python](https://github.com/end-4/dots-hyprland/issues/1017) and [using stable Arch package for Hyprland](https://github.com/end-4/dots-hyprland/issues/1080)), now it installs and runs pretty stable and smooth, so I think it's suitable to revert this PR now.

vaxerski:
that would be up to @end-4

clsty:
> that would be up to @end-4

Yeah that makes sense.

@end-4 you can actually be more confident about your project because many major issues complained by users were actually caused by versioning/installation thingy (which should be my responsibility...) but now they are much stabler essentially.

end-4:
Thank you @clsty for your dedication, and I agree. vaxry, can you revert this?

fufexan:
Reverted.

vaxerski:
thanks fuf

PR #902, opened 2024-12-15
binds: add nested submaps example

davc0n:
ref. https://github.com/hyprwm/Hyprland/issues/8672

### DIFF
--- a/pages/Configuring/Binds.md
+++ b/pages/Configuring/Binds.md
@@ -427,25 +427,25 @@ submap = main_submap
 # ...
 
 # nested_one
-    bind = , 1, submap, nested_one
-    submap = nested_one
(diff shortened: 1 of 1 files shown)


pages/Configuring/Binds.md
```diff
@@ -416,7 +416,43 @@ submap = reset
...
+Submaps can be nested, see the following example:
+
+```ini
+bind = $mainMod, M, submap, main_submap
+submap = main_submap
+
+# ...
+
+# nested_one
+    bind = , 1, submap, nested_one
```
vaxerski, line 430: Don't add those indents, it will give people a weird idea that this is needed

vaxerski (approved):
thanks!

PR #913, opened 2024-12-22
hyprland-qtutils: add disable warning instructions

davc0n:
Saw the warning and searched on google about that.
First result is this wiki page with no information about how to disable the warning... currently it is documented (see the reference), but not immediate to find, that's the reason of this PR.

### DIFF
--- a/pages/Hypr Ecosystem/hyprland-qtutils.md
+++ b/pages/Hypr Ecosystem/hyprland-qtutils.md
@@ -4,3 +4,6 @@ title: hyprland-qtutils
 ---
 
 hyprland-qtutils is a small bunch of utility applications that hyprland might invoke. These are stuff like dialogs or popups.
+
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
you see, I don't like this, because you should install it and not ignore warnings.

davc0n:
I agree that generally users should install it, but haven't noticed any issue so I decided to disable the warning (at least for now).
Who wants to disable the warning knows there's a way to do so, this just reduce searching time.

Eventually we could add a message saying that it's strongly recommended to install that package?

vaxerski:
> haven't noticed any issue

well because it's not present, so hyprland doesn't run it, so it disables features that depend on it. That's why you don't see it.

davc0n:
>stuff like dialogs or popups.

May I ask you some screenshots? Can we add those to the documentation page eventually? Thanks in advance for your support :)

PR #922, opened 2024-12-30
I think you might want to add my dotfiles to wiki

BinaryHarbinger:
My dotfiles has a tons of scripts which provides functionalities like changing wallpapers while using hyprpaper. there is also cafein like script which as provides good functionality to dotfiles.


fufexan:
We're not admitting random repos. The ones already listed have tons of stars and people looking over them.

PR #927, opened 2025-01-04
Clipboard-Managers: remove broken manager

kociap:
The manager `cb` does not support Wayland. The version available in AUR renders the entire desktop unusable by spawning invisible windows which cause focus loss and repositioning of all windows every ~2 seconds. Does not work with `wl-paste --watch` either.


vaxerski (approved):
lgtm

PR #928, opened 2025-01-04
Document that hyprlock config is required.

rcorre:
Without a config, hyprlock fails with:

```
[CRITICAL] ConfigManager threw: Could not find config in HOME, XDG_CONFIG_HOME, XDG_CONFIG_DIRS or /etc/hypr.
```

This was mentioned in https://github.com/hyprwm/hyprlock/issues/450 and closed with

>  hyprlock with no config is more confusing than anything, as it will not render anything

So I'm assuming this is behaving as expected and the docs just need to be fixed.

### DIFF
--- a/pages/Hypr Ecosystem/hyprlock.md
+++ b/pages/Hypr Ecosystem/hyprlock.md
@@ -8,8 +8,7 @@ for Hyprland.
 
 ## Configuration
 
-Configuration is done via the config file at `~/.config/hypr/hyprlock.conf`. It
-is not required, but recommended. Without it, locking shows the current screen.
+Configuration is done via the config file at `~/.config/hypr/hyprlock.conf`. This file must exist to run `hyprlock`.
 
 ### Variable types
 


vaxerski (approved):
yep, thanks!

abctaylor:
Where is an example `~/.config/hypr/hyprlock.conf` provided? Thanks.

abctaylor:
https://github.com/hyprwm/hyprlock/issues/365#issue-2350554844 looks like a good baseline.

rcorre:
At least on Arch, `hyprlock` ships with an example config:

```
❯ cat /usr/share/hypr/hyprlock.conf 
# sample hyprlock.conf
# for more configuration options, refer https://wiki.hyprland.org/Hypr-Ecosystem/hyprlock

input-field {
  monitor =
  fade_on_empty = false
}

background {
  color = rgb(23, 39, 41)
}
```

You actually don't _need_ anything in the config, the file simply needs to exist, but this will show the current screen while locked.

PR #942, opened 2025-01-10
Environment-variables: add QT_QUICK_CONTROLS_STYLE variable

izmyname:
Add QT_QUICK_CONTROLS_STYLE variable under "QT variables" sections.

Also, I'd add this variable to Hyprland example config.


vaxerski:
you dont need this env and shouldnt set it

izmyname:
Indeed, the theming works without the variable. I didn't notice that

PR #944, opened 2025-01-12
Configuring/Multi-GPU: Fixed AQ_DRM_DEVICES examples…

Farshief:
to reflect avoidance of dynamically assigned symlinks as mentioned in the preceding paragraph:
```Do not use the card1 symlink indicated here. It is dynamically assigned at boot and is subject to frequent change, making it unsuitable as a marker for GPU selection.```

### DIFF
--- a/pages/Configuring/Multi-GPU.md
+++ b/pages/Configuring/Multi-GPU.md
@@ -65,7 +65,7 @@ If instead you would like to use another GPU, or the wrong GPU is picked by defa
 set `AQ_DRM_DEVICES` to a `:`-separated list of card paths, e.g.
 
 ```plain
-env = AQ_DRM_DEVICES,/dev/dri/card0:/dev/dri/card1
+env = AQ_DRM_DEVICES,/dev/dri/by-path/pci-0000:01:00.0-card:/dev/dri/by-path/pci-0000:06:00.0-card
 ```
 
 Here, we tell Hyprland to set priorities. If `card0` isn't available for
@@ -80,7 +80,7 @@ including Hyprland, or default to your dGPU if you prefer.
 This method ensures that the variable is properly exported to the systemd environment without conflicting with other compositors or desktop environments.
 
 ```plain
-export AQ_DRM_DEVICES="/dev/dri/card0:/dev/dri/card1"
+export AQ_DRM_DEVICES="/dev/dri/by-path/pci-0000:01:00.0-card:/dev/dri/by-path/pci-0000:06:00.0-card"
 ```
 
 {{< /callout >}}


Farshief:
Now I'm questioning if this change is actually correct or not. I was re-reading the entire page today and realized that while it does say not to use the card symlinks as they may change at boot, it also indicates that the list of cards in `AQ_DRM_DEVICES` should be separated by a `:`.

This seems to preclude the entire idea of using the full by-path card location since the cards themselves contain `:` multiple times in their full names. As a result I'm left not sure what the actual "most correct" answer is.

fufexan:
Indeed, this would garble AQ_DRM_DEVICES. @vaxerski can we change the separator to `,`?

vaxerski:
no

vaxerski:
idk who wrote "it changes", does it? I don't think  it does...?

Farshief:
It appears that the inconsistency was also noticed [here](https://github.com/hyprwm/hyprland-wiki/issues/807) and according to the comments there it's also not possible to escape the `:` in the path.

Also according to that post the symlinks aren't recognized. I'll admit I'm a bit over my head on this so I'll defer to more advanced users on it.

Farshief:
I just wanted to add that I did some testing tonight and the symlinks do function and appear to be the only method to set AQ_DRM_DEVICES since the full card names include `:` and they can't be backslash escaped.

One workaround is to create more permanent symlinks in a separate folder and reference them since the `/dev/dri` symlinks are generated at boot and theoretically could change (Mine seem to stay the same over several boots.)

One thing is clear this PR should not be accepted. I plan to do some more testing tomorrow and submit an updated PR rewording the tip about changing symlinks and adding the workaround advice.

vaxerski:
thanks! 

Generally I don't own or use multi-gpu systems, but card0 and card1 are generally consistent from what I have seen in general? You could also obviously write a script to make a symlink without `:`s e.g. `myEpicCard1` on boot.

GKHWB:
Actually I can confirm that card* symlinks can change on boot, I'm currently struggling with my hyprland config as a result because I'd like to use by-path instead but the `:` separator prevents that.

vaxerski:
why not execute `ln -s /blahblah/card /home/me/cards/myEpicCard0` and use the other path for AQ?

Farshief:
From what I read soft links may not work since they just point to a file path (which would still include the dreaded `:`. So I was experimenting with hard links since it just references the inode number.

Alas I also just updated the system and now it's stuck on `a compositor or graphical-session* target is already active` which is something it usually says but now it's not starting the graphical display so I've had to put my experimenting on hold until I fix that.

GKHWB:
Using a symlink does work, I just think it'd be better to be able to directly specify the card in AQ instead of having to use a kinda messy (and not declarative in NixOS without extra scripts) symlink solution.

If an escape for `:` is for some reason impossible or too difficult to add tho then fair enough.

fufexan:
You can use [`systemd.tmpfiles`](https://search.nixos.org/options?channel=24.11&show=systemd.tmpfiles.rules&from=0&size=50&sort=relevance&type=packages&query=tmpfiles) to declaratively create symlinks.

GKHWB:
Oh, thanks for the tip

Farshief:
It turned out that me trying to use a symlink causes my graphical environment not to even launch and was unrelated to the previous error. I'll continue investigating. 

@GKHWB did you have success booting with symlinks?

GKHWB:
Sorry for the long respond time I kinda forgot, yes @Farshief booting with symlinks worked for me

PR #951, opened 2025-01-20
fix: hyperlink to 404 page had an extra .md at the file name

chedieck:
In https://wiki.hyprland.org/Configuring/Binds/, the hyperlink on the sentence
- "If you want to access your description you can use `hyprctl binds`. For more information have a look at **Using Hyprctl**" 

is as pointing to a non-existent location. That is because it points to https://wiki.hyprland.org/Configuring/Using-hyprctl.md when it should point to https://wiki.hyprland.org/Configuring/Using-hyprctl  (no .md at the end).

### DIFF
--- a/pages/Configuring/Binds.md
+++ b/pages/Configuring/Binds.md
@@ -182,7 +182,7 @@ for example,
 bindd = SUPER, Q, Open my favourite terminal, exec, kitty
 ```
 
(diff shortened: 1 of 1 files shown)


vaxerski (approved):
https://github.com/user-attachments/assets/f0666b4c-39a3-4645-b877-e038d70e9fd9

PR #969, opened 2025-02-08
Update Fedora version numbering from 39 to 40 (as Fedora 39 is EOL now)

Angxddeep:
update,🤷🏻‍♂️


fufexan (approved):
I guess 39+ implies 40 as well but anyway

Angxddeep:
> I guess 39+ implies 40 as well but anyway

Well at any given time there are always 2 current versions that are supported so 39+ implies 39 & 40 and 40+ implies 40 and 41 which are currently both supported

PR #979, opened 2025-02-16
Hypr-Ecosystem/hyprsunset: add 'run at startup' section

izmyname:
Since hyprsunset now can be controlled via IPC, it makes sense to autostart it with the rest of the session. 

Also, https://github.com/hyprwm/hyprsunset/pull/19

### DIFF
--- a/pages/Hypr Ecosystem/hyprsunset.md
+++ b/pages/Hypr Ecosystem/hyprsunset.md
@@ -27,6 +27,10 @@ below the monitor's minimum.
 
 See `hyprsunset --help`.
 
+## Run at startup
+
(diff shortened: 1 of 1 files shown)


pages/Hypr Ecosystem/hyprsunset.md
```diff
@@ -27,6 +27,10 @@ below the monitor's minimum.
 
 See `hyprsunset --help`.
 
+## Run at startup
+
+To autostart hyprsunset, add: `exec-once = hyprsunset` to your `hyprland.conf` . If you start Hyprland with [uwsm](../../Useful-Utilities/Systemd-start), you can also use `systemctl --user enable --now hyprsunset.service` command.
```
fufexan, line 32:
```suggestion
To autostart hyprsunset, add: `exec-once = hyprsunset` to your `hyprland.conf`.
Alternatively, use `systemctl --user enable --now hyprsunset.service` in order to use hyprsunset as a systemd user service.
```

Using the service and using uwsm are orthogonal.

fufexan (approved):

PR #987, opened 2025-02-28
Getting Started/Installation: Adding missing dependencies in manual installation for Arch Linux

abdalrahmanshaban0:


fufexan (approved):

abdalrahmanshaban0:
> Shouldn't it be `hyprland-qtutils-git` ?

It's not necessary at compiling time, but a message appears after installation that you should install hyprland-qtutils.

PR #1000, opened 2025-03-10
Update Systemd-start.md

Vladimir-csp:
- `systemd-cat` no longer needed.
- add more info about autostart.


fufexan (approved):
LGTM, thanks!

PR #1013, opened 2025-03-22
Configuring/Example-configurations: Add an example screenshot

Sly-Harvey:
Made a configuration and thought some people could learn from it, so wanted to share.


fufexan:
We are against adding more configurations here. There are plenty already.

Sly-Harvey:
ok that makes sense, apologies for the pr was unaware that there was too many already.

fufexan:
No problem, it was our fault for not specifying this anywhere. There's now a callout with this decision on the wiki page.

PR #1014, opened 2025-03-23
Changing "debs" to "devs"

Hendriksons:
Changing "debs" to "devs" in "Contibuting and Debugging" since i deem "debs" to be nonsensical in this context


vaxerski:
deps, dependencies

PR #1037, opened 2025-04-16
Alttab

aphelei:
add alt tab behaviour

testing needed please


fufexan (commented):
The disable.sh script is referenced but not provided.

aphelei:
yeah sorry

fufexan (commented):
Generally looks good, but

1. I'd add `#!/usr/bin/env bash` shebangs at the top of scripts
2. You can add \`\`\`bash {filename="alttab.sh"} to also display the filename inside the code block
3. Explicitly list the dependencies.
4. Use `$XDG_CONFIG_HOME` instead of `~/.config`, as it's more idiomatic. Besides, `~` does not always expand to `$HOME` inside scripts.
5. Mention that `foot --server` *has* to be running.

fufexan (approved):
LGTM, thanks!

PR #1042, opened 2025-04-23
Hypr Ecosystem/hyprpaper: added example random wallpaper script for multi-monitor setups

wheelsbot7:
Also formatted the markdown with Prettier, but the hyprpaper change was the only one of substance.


fufexan (commented):
Generally alright, but I'd prefer formatting in a separate commit, and the table should remain unformatted.

wheelsbot7:
Undid formatting with an amend. Now the added text is the only change.

fufexan (commented):
LGTM, thanks!

PR #1064, opened 2025-05-13
Nix/Cachix: improve substituters documentation

FridayFaerie:
For a substituter to be used, [one of the following has to be met](https://nix.dev/manual/nix/2.24/command-ref/conf-file.html#conf-substituters):

- the substituter is under the trusted-substituters list
- The user calling Nix is in the trusted-users list

I think it is better to recommend adding hyprland.cachix.org as a trusted substituter, rather than adding to the trusted-users list, but at least one of these needs to be done for the substituter to be used.


fufexan (approved):
Great, I didn't know about this.
LGTM

PR #1067, opened 2025-05-13
permissions: add plugin path example

sudoAlphaX:
Give an example to always allow loading a plugin from a hardcoded path.

Uses path `/var/cache/hyprpm/$USER/hyprland-plugins/pluginName.so` for demonstration


sudoAlphaX:
hyprpm installed the plugins in `/var/cache//hyprwm/$USER/hyprland-plugins/` in my system.

Let me know if i need to include the `~/.local/share/hyprwm/hyprland-plugins` path also

vaxerski:
but why would you show this when you can just allow hyprpm

sudoAlphaX:
> but why would you show this when you can just allow hyprpm

Useless pr then. My bad. Closing.

PR #1081, opened 2025-05-25
pages/Hypr Ecosystem/hyprcursor.md: added use of hyprcursor-util

johnwill97:
Added instruction on how to use hyprcursor-util,
from looking for cursors on what to look for,
to extracting it using hyprcursor-util,
then lastly to create the cursor still using
hyprcursor-util

### DIFF
--- a/pages/Hypr Ecosystem/hyprcursor.md
+++ b/pages/Hypr Ecosystem/hyprcursor.md
@@ -36,9 +36,51 @@ env = HYPRCURSOR_SIZE,24
 
 ## Creating / Porting Themes
 
-Go to the [hyprcursor repo](https://github.com/hyprwm/hyprcursor)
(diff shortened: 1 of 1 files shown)


vaxerski (commented):
you should not do that unless strictly necessary. It's always better to just remake the theme with svg cursors instead.

vaxerski:
svg are vector graphics. XCursor themes are rasterized. You can go to bibata's source repo and grab the svgs there (though they do put them through a custom renderer first for different themes)

At any rate bibata is already done and you can find it on our discord server or at https://github.com/LOSEARDES77/Bibata-Cursor-hyprcursor

vaxerski:
I just dont think we should encourage this. Instructions to make themes are already detailed in docs/.

johnwill97:
ok, thanks anyway for the responses, i'll close this now

PR #1083, opened 2025-05-25
Hyprpaper: Fix anchor link

Madeleaan:
Fixes anchor link pointing to #IPC which does not exist, instead of #ipc


Madeleaan:
I might be blind and did not notice #1082

PR #1084, opened 2025-05-25
Update Variables.md. Set scroll_button_lock from `0` to true. Literally one line

xNyfPtx:
Set scroll_button_lock from `0` to true


vaxerski (commented):
0 is false.

xNyfPtx:
> 0 is false.

my bad, I assumed that the first one in each comma is true, ex (yes and no) and the second one is false (which is no)
image

xNyfPtx:
i just updated it

vaxerski (approved):

PR #1090, opened 2025-05-26
Variables: explicitly mention the subcategory name for each subcategory

gwennlbh:
I use these to automatically generate data for [HyprLS](https://github.com/hyprland-community/hyprls)

Of course, it's up to me to adapt my kinda-ugly scraping to your changes and not the other way around, but I figured It'd be easier for me to add these everywhere upstream instead of like, trying to infer the group name from the heading or something


fufexan (approved):
LGTM, thanks!

PR #1103, opened 2025-06-01
docs: Improve docs regarding Caps Lock remapping

New9c:
I want people to know that many options exist for Caps Lock in `kb_options`, so I made the docs regarding Caps Lock a bit clearer.


fufexan (approved):
Great, LGTM!

PR #1119, opened 2025-06-15
fix path to `Variables` page

scifisatan:
The `Variables` link on `Start` page leads to 404, as the Start page and Variables page are on same dir, the reference `../Variables` won't work. Simply removing `.` would make the path correct


scifisatan:
i made another pr #1120  that fixes all the paths in `Start` page.

PR #1120, opened 2025-06-15
Path fixes on `Start` page

scifisatan:
As `Start` page and pages it references like `Keywords` & `Variables` are on same directory, `./` should be used.


fufexan:
Okay, I can repro this. The real problem is the link on https://wiki.hypr.land/Configuring/. It should have a `/` at the end.

scifisatan:
wow nice catch.

PR #1122, opened 2025-06-16
Fix link to Useful Utilities

Fittiboy:
Without the trailing `/`, all links under "Useful Utilities" lead to a 404 page.


fufexan (approved):
LGTM, thanks!

vaxerski:
@fufexan apparently
```
markup:
  goldmark:
    renderHooks:
      link:
        enableDefault: true
```

then we can just `[link](page.md)`, can you test it and maybe change up the links so that this can stop happening?

PR #112, opened 2023-02-05
Added fix instructions for hybrid graphics devices

MathisP75:
I did just what I added on top of the other instructions for Nvidia and it worked for me. Could be very useful to others.

PR #1124, opened 2025-06-18
nvidia: change wording for multi-monitor with hybrid graphics

aomiki:
It feels weird to see this section, since it implies that multi-monitor doesn't work on hybrid graphics. 

But I'm writing this PR from a laptop with hybrid graphics (nvidia+ intel) and 2 monitors connected.
In fact, I can't disable hybrid since one of the monitors is connected through thunderbolt which wouldn't function without iGPU.

So TL;DR:
Multi-monitor on hybrid graphics works perfectly fine. Reworded a bit, so it doesn't imply otherwise. Maybe better to remove the section altogether? 
It was originally added in #112, but it's not clear how my setup is different, except that 2 years have passed since.

### DIFF
--- a/pages/Nvidia/_index.md
+++ b/pages/Nvidia/_index.md
@@ -208,11 +208,12 @@ experimental support for Chromium, however there has not been much success.
 ## Other issues
 
 ### Multi-monitor with hybrid graphics
+If you experience issues with multi-monitor setup on a hybrid graphics device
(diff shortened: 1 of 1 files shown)


fufexan (approved):
LGTM

I don't think any of the HyprWM members have hybrid setups, so we haven't tested this.

PR #1128, opened 2025-06-22
Configuring/Permissions: Add docs for NixOS paths

dasJ:


fufexan (commented):
I use string interpolation in my nix-generated config, so I would add that that approach also works.

fufexan (approved):
LGTM, thanks!

PR #1131, opened 2025-06-27
Update _index.md

bg-l2norm:
nvidia: added  workaround for Electron/Chromium boot delay on hybrid GPUs


bg-l2norm:
Callout for a known issue where Electron or Chromium apps stall after boot on setups with an Intel iGPU and Nvidia dGPU. Solution involves module ordering in mkinitcpio.conf.

fufexan (approved):
LGTM, thanks!

PR #1132, opened 2025-06-27
Remove spaces in rgba and rgb decimal value's and added explanation

Pucas01:
Hyprland does not work with the decimal version of rgba() and rgb() when they have spaces inbetween the numbers, so here a change to the wiki to explain this as its a little confusing at the moment, maybe im stupid but i feel like this is a good change (don't mind the commit descriptions)

### DIFF
--- a/pages/Configuring/Variables.md
+++ b/pages/Configuring/Variables.md
@@ -34,6 +34,7 @@ You have 3 options:
 
 rgba(), e.g. `rgba(b3ff1aee)`, or the decimal equivalent `rgba(179,255,26,0.933)`
 (decimal rgba/rgb values should have no spaces between numbers)
+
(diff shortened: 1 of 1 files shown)


fufexan:
This is the right way to go. However there are still occurrences of rgba with spaces in the wiki that need to be fixed.

Pucas01:
After using grep to look for all the rgba/rgb's used in the wiki i have found that the only other time it gets mentioned is in the hyprlock page, as shown here:

2025-06-28-114347_hyprshot

i tested it and turns out hyprlock works with both rgba(0, 0, 0, 0) and rgba(0,0,0,0)

so should these keep the spaces or should they also be removed for consistency?

fufexan:
They can be left as they are now. Thanks for looking into it.

PR #1133, opened 2025-06-27
UWSM: Fix hyprland's UWSM entry name

Lem0nbleach:
Not sure if this is only on my system or true for everyone, but hyprland.desktop does not exist on my computer and the correct entry is hyprland-uwsm.desktop.

Thus creating this pull request in case others might get locked out by their shell profile because of this.


fufexan (approved):
LGTM, thanks!

PR #1143, opened 2025-07-11
Nix/Hyprland-on-Home-Manager: add information on how to export variables when using NixOS with UWSM

zeorin:
When using the NixOS module with UWSM, one should not enable the Home Manager module's systemd integration, as it conflicts with UWSM. Thus, configuring `wayland.windowManager.hyprland.systemd.variables` will not work, as it only takes effect when `wayland.windowMarager.hyprland.systemd.enable == true`.

To ensure that the environment variables for user services are set correctly even if they start before Hyprland is started by UWSM, `${XDG_CONFIG_HOME}/uwsm/env` should be populated. Since UWSM sources that file, we can use the session variables script that Home Manager generates under the hood verbatim.


fufexan (approved):
LGTM

PR #1162, opened 2025-07-20
Configuring and Getting Started/index: fold by default

### LINKED ISSUE #826: when clicking on a category other categories unfold
so say i fold every category:
image
and then i decide i want to click on a category:
image
the two categories i folded previously: automatically unfold, this needs to be fixed ASAP

Bugg4:
This sets both categories to start as folded by default, eliminating the unwanted behavior of them unfolding every time the user clicks on a category.

Resolves #826

### DIFF
--- a/pages/Configuring/_index.md
+++ b/pages/Configuring/_index.md
@@ -1,8 +1,6 @@
 ---
 weight: 3
 title: Configuring
-sidebar:
-  open: true
 ---
 
 This section is all about the configuring (aka ricing) of your Hyprland experience.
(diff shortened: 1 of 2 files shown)


fufexan:
I'd be in favor of this, but we decided to keep them open some time ago so that new users can easily find things.

@vaxerski have you changed your mind in the meantime? I don't think it really affects page visibility.

Bugg4:
> I'd be in favor of this, but we decided to keep them open some time ago so that new users can easily find things.

I figured it was done for that reason, and as I stated in my last comment in #826 it would have been totally ok, if only the `open` property din't re-apply for every page load.
Until the behavior isn't fixed by Hextra, imo, in its current state, it annoys users more than it helps them, because once users figure out they can unfold entries in the sidebar (which hopefully doesn't take that much effort), then the "always open" behaviror isn't really needed. My 2 cents

PR #1177, opened 2025-07-26
Add section about udev rule for consitent DRM card device paths

m-bartlett:
Mentions a solution for the issue of `/dev/dri/card*` paths changing periodically which GPU they point to using a udev rule, which also retains the ability to specify multiple cards in `AQ_DRM_DEVICES`.


fufexan (approved):
LGTM, didn't know about this.

PR #1184, opened 2025-08-03
Options from hyprwm/Hyprland#11214

xzn:
Hope I'm doing this correctly, I don't see version tags on this repo.


xzn:
For hyprwm/Hyprland#11214

PR #1185, opened 2025-08-04
Backend: update favicons and manifest

### LINKED ISSUE #1179: Proposal for slight improvements to the Hyprland logo
Looking at the svg file for the Hyprland logo, I noticed that the nodes are a bit wonky, and there are definitely more than what's necessary.

Image

My guess is that happened because the logo was either hand-trace from a raster version, or automatically vectorized. Probably the latter.
So tried to recreate it with less nodes.

Image

It's note a 100% perfect copy, because it's kinda hard to discern which changes in thickness were actually intended, and which were due to the way it's been vectorized.
Here's a [comparison](https://slow.pics/c/dHQjC33i):

The new version is a tad thicker, because it looks good to me, but I'm obviously open to feedback so we can get it exactly how _you_ like it :)

Bugg4:
- Update master svg for Hyprland logo
- Rerender favicons based on new master
- Set dark theme variant for navbar logo
- Fill missing fields in webmanifest
  - Change theme and background color to `#111111`

New renders are a slightly larger, ~but still respect the safe area~ specified by the [Google Play Icon Design Specification](https://developer.android.com/distribute/google-play/resources/icon-design-specifications#attributes). 
Edit: apparelty they don't because they still get cut off if `maskable` is added to the `purpose` key in manifest.
So I removed `maskable` for now. I'll look into full PWA icons support at a later time.

I also took a liberty on the short name in manifest: "HyprWiki".
This is basically the app name that appears on Android when the wiki is saved as a PWA. 

Resolves #1179


Bugg4:
This fixes it, but I'm not quite sure what was wrong with it.
```toml
path = "favicon.svg"
dark = "favicon-dark.svg"
```
Resulted in the light theme navbar icon being smaller, even if they're the same exact file, just renamed.

```
path = "favicon.svg"
```
Resulted in both light and dark being too small.

```
path = "favicon-dark.svg"
```
Show the correct size on both themes. Not sure why this happens tho

Bugg4:
Turns out I had to explicitly clear the browser's cache because the `--noHTTPCache` wasn't enough tu fully refresh the build.
`hugo serve --ignoreCache --noHTTPCache` does appear to achieve a full and clean refresh.
I'll revert the previous fix.

PR #1186, opened 2025-08-04
Configuring/Binds: Improvements, punctuation, clearer wording

Bugg4:
- Enclose some notes in the appropriate callout type
- Add AZERTY link
- Tabularize bind flags
- Rearrange the Bind flags section to appear before they're used in other examples
- Remove duplicate Binding mods section

rebase of #1158

### DIFF
--- a/content/Configuring/Binds.md
+++ b/content/Configuring/Binds.md
@@ -38,9 +38,6 @@ See the
 [xkbcommon-keysyms.h header](https://github.com/xkbcommon/libxkbcommon/blob/master/include/xkbcommon/xkbcommon-keysyms.h)
 for all the keysyms. The name you should use is the segment after `XKB_KEY_`.
 
-If you are unsure of what your key's name is, you can use `xev` or `wev` to find
-that information.
-
 If you want to bind by a keycode, you can put it in the KEY position with
 a `code:` prefix, e.g.:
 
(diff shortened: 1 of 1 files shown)


content/Configuring/Binds.md
```diff
@@ -64,55 +67,102 @@ To get the correct name for an `unmodified_key`, refer to [the section on uncomm
...
+| `e`  | repeat          | Will repeat when held.                                                                                       |
+| `n`  | non-consuming   | Key/mouse events will be passed to the active window in addition to triggering the dispatcher.               |
+| `m`  | mouse           | See the dedicated [Mouse Binds](#mouse-binds) section.                                                       |
+| `t`  | transparent     | Cannot be shadowed by other binds.                                                                           |
+| `i`  | ignore mods     | Will ignore modifiers.                                                                                       |
+| `s`  | separate        | Will arbitrarily combine keys between each mod/key, see [Keysym combos](#keysym-combos).                     |
+| `d`  | has description | Will allow you to write a description for your bind.                                                         |
+| `p`  | bypass          | Bypasses the app's requests to inhibit keybinds.                                                             |
```
fufexan, line 123: Can you remove the extra spaces? I know it's autoformatted but this generates huge diffs when line lengths change.

Bugg4: You mean going from:
```| `l`  | locked          | Will also work when an input inhibitor (e.g. a lockscreen) is active.                                        |```
to:
```| `l` | locked | Will also work when an input inhibitor (e.g. a lockscreen) is active. |```
for all rows, correct?.
Should I keep all tables unformatted going forward as a rule of thumb, even smaller ones?

fufexan: Yeah. I wonder if we could get a prettierrc for this, maybe even a dinit rule.

PR #1191, opened 2025-08-08
Configuring/Dwindle Layout: Improvements, punctuation, clearer wording

Bugg4:
- Turn `Quirks` into a `Preface` section
- Improve formatting for table content
- mouse -> cursor

A couple of clarifying questions:
- `single_window_aspect_ratio_tolerance` has a type of int, but a default value of `0.1` and a hint for `0 - 1` range? Should it be a float?
- `permanent_direction_override`: did I understand correctly? The possible values are either `true`, `false`, `left`, `right`, `top`, `bottom` or any other non-direction string?
- in `split bias` I'm not sure I get how `positional` behaves

### DIFF
--- a/content/Configuring/Dwindle-Layout.md
+++ b/content/Configuring/Dwindle-Layout.md
@@ -46,7 +46,7 @@ category name: `dwindle`
 | `split_width_multiplier` | Specifies the auto-split width multiplier. <br> Multiplying window size is useful on widescreen monitors where window W > H even after several splits. | float | `1.0` |
 | `use_active_for_splits` | Whether to prefer the active window or the cursor position for splits. | bool | `true` |
 | `default_split_ratio` | The default split ratio on window open. <br> `1.0` means an even 50/50 split. | float [`0.1 .. 1.9`] | `1.0` |
-| `split_bias` | Specifies which window will receive the larger half of a split. <br> `0` -> positional <br> `1` -> current window <br> `2` -> new window | int [`0\|1\|2`] | `0` |
+| `split_bias` | Specifies which window will receive the larger half of a split. <br> `0` -> positional: the new window is larger only if it opens to the `top` or `left` of the parent. <br> `1` -> always the current window. <br> `2` -> always the new window. | int [`0\|1\|2`] | `0` |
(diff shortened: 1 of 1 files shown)


vaxerski:
1. yes

the rest are community-written and I don't use them. Might be worth looking into the code

Bugg4:
- Looks like `permanent_direction_override` takes an `int`, but in the code is used basically as bool: [DwindleLayout.cpp](https://github.com/hyprwm/Hyprland/blob/aaedce596ec27742ea8f00a20607913e8a3e83db/src/layout/DwindleLayout.cpp#L348)
@DashieTM do you happen to remember the intended reason for it being an `int` ? Can it be made a `bool` instead?

- @myQwil can you please clarify on what `positional` means?

Bugg4:
> @Bugg4 while I am no longer sure why it was and still is an int in code, the use case is definitely just a boolean flag. My assumption is that I modeled it after what existed, like smart split, which has an int in code but is a boolean in the wiki.

Thank you! Yea, taking a quick look at [config.cpp](https://github.com/hyprwm/hyprlang/blob/23f0debd2003f17bd65f851cd3f930cff8a8c809/src/config.cpp#L267) from Hyprlang, it seems that internally "true", "yes" and "on" are interpreted as `int`s since we don't have the concept of a boolean config value ([config.hpp](https://github.com/hyprwm/hyprlang/blob/23f0debd2003f17bd65f851cd3f930cff8a8c809/src/config.hpp#L33-L40))

Bugg4:
> The opening window will be the bigger window if it opens above or to the left of the current window, and smaller if it opens below or to the right of the current window.

Ah gotcha, that's the piece I was missing: top/left takes precedence over bottom/right. Tanks for the clarification 👌🏻

myQwil:
This got me thinking about the `split_bias` variable. It can probably be simplified to just two options instead of three, in which case, it could probably just be a boolean.

It could be changed to something like `directional_split`. If true, the split bias is direction-based, and if false, it's window-based.

The reason we don't need a third option is because we can simply invert the ratio to give the bias to the opposite window or the opposite directions.

PR #1197, opened 2025-08-11

jgmagift:
I added gb layout type for those who uses UK layout, this helps users to know more variations of keyboard types.


fufexan:
I don't think this is necessary. https://wiki.hypr.land/Configuring/Variables/#xkb-settings instructs users where to look to find more layouts.

PR #1211, opened 2025-09-01
Talk about private visibility in Advanced.md

jmanc3:


vaxerski (commented):
needs a mention that some STL headers will break if this is done, and a workaround of including them beforehand

jmanc3:
Note that some Hyprland files include STL files which may end up breaking if you do this. If you encounter this issue, include STL file before the section where include the Hyprland file.

Something like that?

vaxerski (approved):

PR #1218, opened 2025-09-07
docs(binds): persistence warning for hyprctl

sand4rt:
partial fix for: https://github.com/hyprwm/Hyprland/discussions/11583#discussioncomment-14331906

### DIFF
--- a/content/Configuring/Binds.md
+++ b/content/Configuring/Binds.md
@@ -212,6 +212,13 @@ bindl = , switch:on:[switch name], exec, hyprctl keyword monitor "eDP-1, disable
 bindl = , switch:off:[switch name], exec, hyprctl keyword monitor "eDP-1, 2560x1600, 0x0, 1"
 ```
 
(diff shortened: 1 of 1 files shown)


content/Configuring/Binds.md
```diff
@@ -212,6 +212,13 @@ bindl = , switch:on:[switch name], exec, hyprctl keyword monitor "eDP-1, disable
 bindl = , switch:off:[switch name], exec, hyprctl keyword monitor "eDP-1, 2560x1600, 0x0, 1"
 ```
 
+{{< callout type=warning >}}
+
+Reloading the config removes `hyprctl keyword` commands set in the switches. To make them persistent, add:
+`exec = sh -c 'if [ $(hyprctl monitors | grep -c "Monitor") -gt 1 ]; then hyprctl keyword monitor "eDP-1, disable"; fi'`
```
fufexan, line 218: This should be in a codeblock (\`\`\`ini)

PR #1219, opened 2025-09-07
Fix TOC on a few pages via markdown headings

rmbruntz:
Hextra doesn't put H1 headings in the TOC / right sidebar. This `#`-indents every heading on these pages by 1 so all of them show up.

### DIFF
--- a/content/Contributing and Debugging/_index.md
+++ b/content/Contributing and Debugging/_index.md
@@ -3,20 +3,18 @@ weight: 13
 title: Contributing and Debugging
 ---
 
-## Contributing guidelines
-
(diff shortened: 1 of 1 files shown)


content/Contributing and Debugging/_index.md
```diff
@@ -3,20 +3,20 @@ weight: 13
 title: Contributing and Debugging
 ---
 
-# Contributing guidelines
+## Contributing guidelines
```
fufexan, line 6: This could simply be removed. From what I can tell it's the only h1 in the entire file. With this there's no need to change the other headings.

PR #1220, opened 2025-09-08
Page: Update the link to the plugins page on Using-Plugins.md

definite-d:
Updated the link to the Hyprland Plugins page from the old domain (hyprland.org/plugins) to the new one (hypr.land/plugins).

It's not much, but it's honest work.


fufexan (approved):
Makes sense, LGTM

PR #1224, opened 2025-09-10
Useful Utilities/Clipboard-Managers: add `clipvault`

Rolv-Apneseth:
No hard feelings if this isn't wanted, but I made an [alternative](https://github.com/Rolv-Apneseth/clipvault) to `cliphist` and thought it would be cool to include it here, as this is where I originally found `cliphist` too. Thank you all for the great resource that is the Hyprland wiki.


fufexan (approved):
Cool, LGTM

PR #1226, opened 2025-09-12
Warning on "initial" matching

birkirb:
Not sure if this useful but I did have to do quite a bit of debugging to figure why those rules weren't being applied, so perhaps this will save someone that effort or inspire a clarification somewhere else in the documentation.


birkirb:
Hmm.. it doesn't imply you can't match explicitly on them though.

How about I edit that part into something like this?

> Static rules are evaluated once when the window is opened and never again. This essentially means that it is always the `initialTitle` and `initialClass` which must then be matched via `title` and `class`, respectively.

PR #1240, opened 2025-09-27
Update domain for version selector

yechielw:
Openinig version selection leads to the old wiki domain that is blocked by cloudflare (in my case)
image


fufexan (approved):
Thanks, missed that.

PR #1245, opened 2025-10-03
Nix Flake instructions: suggest also adding `package` definition to Home Manager module

non-bin:


non-bin:
> Is your commit incomplete maybe?

no?

fufexan:
Okay, thought you meant something else by this. The info you want to provide already exists in the Hyprland on HM page, but let's add it here as well for visibility.

PR #1258, opened 2025-10-19
Hyprpaper: Add warning about using correct hyprlang syntax when interacting via IPC

Bugg4:
- Add warning about using correct hyprlang syntax when interacting via shell/IPC (related: https://github.com/hyprwm/hyprpaper/issues/282#issuecomment-3419910723)
- Add table with list of keywords, description and params, also serves as a quick summary of available commands and the args they accept.
- Make a subsection for each keyword for clarity
- Titlecase all headings
- Make warning about absolute paths into a warning callout
- Minor formatting


fufexan (approved):
Cool, LGTM

PR #1263, opened 2025-10-23
Update Debian install instructions, not in Trixie

samueloph:
Hyprland was removed from Debian Trixie and Debian Testing due to the
maintainter's request:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1107152

It is only available in Unstable and sadly not up-to-date, since it
can't migrate to Testing and Stable as per the request made.


samueloph:
Oh, my bad, it wasn't a request from upstream, it was the Debian maintainer of the package who requested it.

PR #1264, opened 2025-10-24
Getting-Started/Installation: Remove chaotic AUR mention

vdawg-git:
When there is an ABI break in a .so dependency Chaotic Aur will not rebuild dependent packages.

That causes problems.
A solution would be to bump the minor version of all packages if there is an ABI break, but we dont do that currently.

### DIFF
--- a/content/Getting Started/Installation.md
+++ b/content/Getting Started/Installation.md
@@ -71,10 +71,19 @@ Alternatively, install the `hyprland-meta` package to automatically fetch and co
 ```shell
 yay -S hyprland-meta-git
 ```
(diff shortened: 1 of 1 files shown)


vdawg-git:
Its also not the case for -git in general where the user builds it.
But yeah, true, users can just rely on the official repos.

Edit: But then many plugins dont work (unless they would be maintained by the official repo too). So this is also not optimal.

vdawg-git:
I added a general callout for -git, as it is overall useful to know that updating dependencies might require you to recompile everything.

PR #1276, opened 2025-11-05
Update NixOS instructions for using home-manager with NixOS

urban-prah:
Setting home-manager packages to `null` causes hyprsunset build to fail. The proposed changes suggest better practices and prevent [`hyprsunset/#96`](https://github.com/hyprwm/hyprsunset/issues/69)


content/Nix/Hyprland on Home Manager.md
```diff
@@ -204,18 +204,27 @@ Example configuration:
...
 }
 ```
+You need to make sure to make your system configuration available in home-manager:
+```nix {filename="system.nix"}
+home-manager = {
+  extraSpecialArgs = {
+    hostConfig = config;
+  };
+};
+```
```
fufexan: This is not necessary. When used as a NixOS module, HM will automatically populate `osConfig` with the host `config`.

PR #1280, opened 2025-11-14
Getting Started/Installation: remove outdated notice for Debian

carlo711:
Removed the outdated notice for Debian, as currently hyprland 0.51.1[1] is available on Sid.

[1] https://tracker.debian.org/pkg/hyprland


vaxerski:
heh I wonder if I'll have to reintroduce it soon

PR #1284, opened 2025-11-16
Fix Polish translation for anime girls count

Skidamek:
Original code only handled singular vs plural, failing for counts 2-4 and numbers ending in 2-4 except for teens (12-14). Updated logic correctly handles all positive numbers, including exceptions, ensuring proper grammar: "dziewczynkę", "dziewczynki", or "dziewczynek".

For example `peopleAmount` of `2`, `3`, `4` or `23` should return `Mam 23 dziewczynki anime.` same with e.g. `534` but not if it ends with a teen number like `14` with e.g. `414` the proper form is `Mam 414 dziewczynek anime.`.


vaxerski:
just an example lole

PR #1305, opened 2025-12-04
Update Fedora installation instructions in Installation.md

bowlofneighs:
Updated installation instructions regarding Fedora 43 as Hyprland has been removed from the Fedora repositories


bowlofneighs:
Hi, I think I stumbled upon [this](https://copr.fedorainfracloud.org/coprs/sdegler/hyprland/) up to date `copr` that was forked off of `solopasha/hyprland` that should work

vaxerski:
the problem is I don't really trust random repos...

PR #1306, opened 2025-12-04
Configuring/Uncommon-tips-&-tricks: Improved User-Experience of "Alt tab behaviour"

quantumfate:
The previous implementation executed the `focuswindow` and `alterzorder` dispatchers before the footclient process terminated. This created a timing issue that is particularly noticable during window focusing across workspaces.

For example: A user on workspace 2 uses the alt-tab command to focus one out of several windows on workspace 1 which was not previously focused when the user was working on workspace 1. The focus is correctly updated but Hyprland fails to draw the active border around the newly focused window.

This is fixed by updating the focus after the footclient has terminated.

Detailed changes:
- make the scripts work with resources in a /tmp directory (this is cleaner in oppose the previous approach, especially on NixOS where the directory might be write protected)
- decoupled the focus window logic
- made `alttab.sh` handle the submap bindings
- removed noise from `chafa` parsed images
- explicitely sourced the `foot.ini` to the server command or else the `footclient` won't have user defined styles and defaults to foot's out of the box style

### DIFF
--- a/content/Configuring/Uncommon-tips-&-tricks.md
+++ b/content/Configuring/Uncommon-tips-&-tricks.md
@@ -338,7 +338,7 @@ address=$(hyprctl -j clients | jq -r 'sort_by(.focusHistoryID) | .[] | select(.w
 	      awk -F"\t" '{print $1}')
 
 if [ -n "$address" ] ; then
-	echo "$address" > /tmp/alttab/address
+	echo "$address" > $XDG_RUNTIME_DIR/hypr/alttab/address
 fi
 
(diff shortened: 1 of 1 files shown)


quantumfate:
> > * make the scripts work with resources in a /tmp directory (this is cleaner in oppose the previous approach, especially on NixOS where the directory might be write protected)
> 
> Good idea, but I propose using $XDG_RUNTIME_DIR instead, in case multiple users are running Hyprland with alttab at the same time.

makes a lot of sense! I've changed it.

PR #1307, opened 2025-12-05
Hypr-Ecosystem/hyprpaper: Fix wrong config option (file -> path), increase clarity

zjeffer:
This caused me quite some headaches

Also added a table with the possible wallpaper options, and increased clarity on how the misc section works.


vaxerski (approved):
oops

PR #1329, opened 2025-12-29
Migration script for updating from v0.52

iamkarlson:
Breaking changes is not cool, and figuring out what's wrong just before new year is not cool too. This snippet should probably help 95% of the users to fix their setup right away after the update


vaxerski:
it's unfortunately far from correct or complete, but check https://forum.hypr.land/t/0-53-window-layerrule-converter/1243

PR #1340, opened 2026-01-05
content/breaking-changes: add breaking changes

vdawg-git:
Too many people had a hard time updating to `0.53`. Lets make it a bit easier

### DIFF
--- a/content/breaking-changes.md
+++ b/content/breaking-changes.md
@@ -1,0 +1,85 @@
+---
+title: Breaking changes
+weight: 1
+---
+
+## 10.04.2025
+
+with https://github.com/hyprwm/Hyprland/pull/9945 `cursor:warp_on_change_workspace` no longer controls warps for special workspaces. Use  `warp_on_toggle_special` for more fine control.
+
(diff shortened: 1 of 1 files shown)


vdawg-git:
The problem is that Hyprland now also has downstream consumers who provide a fully configured DE-experience.
That is awesome as it helps a lot of users to transition to the "Full-Linux"-experience (like PewdiePie), however that also means that they dont really know what is configured.
So just a small note in the release-notes will be a bit cryptic. Providing more elaborate instructions does definitely help.
Also we have no control over when someone updates. Maybe someone tries Linux, but doesnt use it, boots again after a couple of months and is greeted with a ton of config errors after updating.
Which also means that tools which help you to migrate like https://itsohen.github.io/hyprrulefix/ stay relevant for quite some timej (I just updated today).
Btw, Vaxry added this tool to the release notes. But imo he shouldnt have to do all that stuff by himself. Let the community help.

To end: Sure we could argue that it is ultimately the users fault for choosing something advanced like Hyprland, but we can also take a step back and view it from the above: People are migrating from Windows to Linux and we should support them.

PR #1347, opened 2026-01-14
Hypr-Ecosystem/hyprlock: add locked bind flag hint to keybinds section

ykahveci:
Adds a note to the hyprlock page, promoting the use of the bind flag `l` (locked) on specific hyprland keybinds to allow those keybinds on the lock screen.

While the bind flags are well-documented on the binds flag already, I think it's helpful to reference this on the hyprlock page as well. (This issue hyprwm/hyprlock#479 seems to confirm that it's not the most obvious feature)


vaxerski (approved):
sure

PR #1370, opened 2026-02-05
fix: add a newline after details closing tag

acropolis914:
Getting Started / Installation.md
- Added a newline after details closing tag. (Fixes instances like trying to render it with other parsers like remark)

Did it so it parses well on my single page application.

Without fix:
image

With Fix:

image


acropolis914:
yay thanks <3

PR #1378, opened 2026-02-14
Use system theme by default

### LINKED ISSUE #1377: Add Theme = Device Default for https://wiki.hypr.land
Currently there is only 'Light' and 'Dark', can you please add optiont for 'Device Default' so that the webpage picks theme from the browser.

mart-e:
Instead of defaulting to dark for those using light system theme

Fixes #1377

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -91,7 +91,7 @@ dark = "favicon-dark.svg"
 width = "full"
 
 [params.theme]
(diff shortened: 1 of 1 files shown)


mart-e:
Oh thanks, I hadn’t figured out yet why my initial PR did not work

PR #1383, opened 2026-02-19
Update hyprpolkitagent.md

heinoscott:
Hypr Ecosystem/Hyprpolkitagent : fixed other possible path

fixed other possible path to '/usr/lib/hyprpolkitagent/hyprpolkitagent' after discovering that the agent runs in the folder with the same name under /usr/lib/. Just including the former path in the hyprland.conf would not start the agent on startup.


fufexan:
What distro are you on?

heinoscott:
> What distro are you on?

CachyOS with kernel version 6.19.2-2

PR #1433, opened 2026-04-18
Configuring/Keywords: update the list of unsupported per-device options

kolayne:
The list of options from the `input {}` block not supported by the `device {}` block was outdated. Added more options and the reason for why they are not supported.

IMO, grouping options configuring devices with options configuring window management is not the right design, so maybe the config should be restructured in the future. For now, just updating the docs.


kolayne:
@vaxerski, ping

vaxerski (approved):

PR #1439, opened 2026-04-26
Configuring/Layouts/Dwindle-Layout: remove pseudotile from example

chocoblocko9:
pseudotile is no longer real


vaxerski (approved):
mandela effekt

PR #1447, opened 2026-04-29
Configuring/Window Rules: Fix syntax in example expressions for size parameter

Deckloins:
The syntax shown in the example doesn't actually work


vaxerski:
this one wont work even more

Deckloins:
That's the way I got it to work. What's the intended syntax ?

PR #1457, opened 2026-04-30
Another batch of random fixes

chocoblocko9:
will keep draft until i find more stuff


chocoblocko9:
@vaxerski could we get a ruling on how to format individual options? on the xwayland page rn i have `xwayland = { force_zero_scaling = true }` which is accurate but kinda ugly, also have `xwayland:create_abstract_socket` which is kinda weird to write for a lua wiki but also used a lot in Variables.md

could be `xwayland.create_abstract_socket` too, think u used that in Variables.md for the categories. i dont really mind but consistency good

chocoblocko9:
also will probably remove the "workspace_swipe, workspace_swipe_fingers and workspace_swipe_min_fingers were removed in favor of the new gestures system..." note in Variables.md cus that happened like literally 8 or 9 months ago, but ill change to lua if its still  wanted

ashley-paul-12345:
hi, while you are at it, can u add this entry too? called `skip_non_kms_dmabuf_formats`
https://github.com/hyprwm/Hyprland/blob/main/src/config/values/ConfigValues.cpp

it's not in quirks in the wiki
https://wiki.hypr.land/Configuring/Basics/Variables/#quirks

chocoblocko9:
> hi, while you are at it, can u add this entry too? called `skip_non_kms_dmabuf_formats` https://github.com/hyprwm/Hyprland/blob/main/src/config/values/ConfigValues.cpp
> 
> it's not in quirks in the wiki https://wiki.hypr.land/Configuring/Basics/Variables/#quirks

oh yeah, that pr just got missed somehow

chocoblocko9:
Hyprlock is tricky, it links back to hyprland a couple times. I rewrote the animations section and linked back to the Blur part of Variables.md but I'm just now realising the variables names don't really line up anymore so we could just link back to the 0.54 pages for those, might be a better idea

Honkazel:
So, if you're fixing random stuff - then look at FAQ. Spoilers: there's still hyprlang config stuff, ~~four if not more broken redirect links~~(fixed in main) and bash scripts, that could be replaced with lua one. That's what I saw
And also Nvidia one. hyprlang config stuff too.

chocoblocko9:
i think like 98% of this is fixed so  ill close it.

PR #1465, opened 2026-05-02
Configuring/Advanced and Cool/Expanding functionality: note the use of LSP for classes and parameters of those classes for Convenience functions

erstarr:
note that the user should use  LSP for classes and parameters of those classes for Convenience functions

### DIFF
--- a/content/Configuring/Advanced and Cool/Expanding-functionality.md
+++ b/content/Configuring/Advanced and Cool/Expanding-functionality.md
@@ -94,6 +94,7 @@ Hyprland exposes a bunch of convenience functions:
  - `hl.exec_cmd()`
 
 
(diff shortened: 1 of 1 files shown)


erstarr:
@vaxerski please merge this before any commits that add params to these classes so it doesn't go stale immediately

vaxerski:
this has been discussed and I don't like it - it _will_ go stale. Why not tell people to use the LSP?

erstarr:
Missed the discussion. LSP is fine. Having these in the wiki all documented in one page is neat, but yeah maintenance burden.

I'll note that people should use the LSP in their code editors for the classes and their params then.

vaxerski (approved):

PR #1496, opened 2026-05-14
Revise Void Linux installation instructions

rXelelo:
Updated third party repository links and installation instructions for Void Linux.

2026-05-14-134210_hyprshot


rXelelo:
The Makrennel/hyprland-void was last time updated year ago and stay on 0.49 i update it to 0.55 and fix issue to make compatible with void

vaxerski:
I'd probably drop the instructions altogether, I can't recommend random third-party repos

PR #1497, opened 2026-05-14
Fix incorrect syntax in global binds section

### LINKED ISSUE #1485: Invalid syntax in global keybinds
For classic-style global keybinds (passing the key events to an application), the hyprland wiki shows the following example (taken from [here](https://wiki.hypr.land/Configuring/Basics/Binds/#classic)):

```lua
hl.bind("SUPER + F10", hl.dsp.pass("class:^(com\.obsproject\.Studio)$"))
```

This has two issues:

1. For this to be valid Lua syntax, the black-slash needs to be escaped `\\`.

2. This throws an error:
   ```
   -C[C]:-1: hl.pass: expected a table { window }
   /home/strokk/.config/hypr/hyprland.lua:176: hl.bind: dispatcher must be a dispatcher (e.g. hl.dsp.window.close()) or a lua function
   ```

Strokkur424:
fixes #1485.

I realized after making a PR to a different repository fixing Lua stuff that the actual syntax was not exactly hard to come by, and having gained that knowledge decided to open a PR for my original issue.

This fixes the argument to the `hl.dsp.pass` dispatcher function by providing a table with a window parameter instead of a plain string. This also fixes the previously incorrectly escaped dot-character in the regex string.

### DIFF
--- a/content/Configuring/Basics/Binds.md
+++ b/content/Configuring/Basics/Binds.md
@@ -274,7 +274,7 @@ Let's take OBS as an example: the "Start/Stop Recording" keybind is set to
 <key>SUPER</key> + <key>F10</key>, to make it work globally, simply add:
 
 ```lua
-hl.bind("SUPER + F10", hl.dsp.pass("class:^(com\.obsproject\.Studio)$"))
+hl.bind("SUPER + F10", hl.dsp.pass({ window = "class:^(com\\.obsproject\\.Studio)$" }))
 ```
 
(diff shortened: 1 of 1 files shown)


morgray-the-dark:
Hello, this should already be fixed in PR #1487. I've noticed the same thing while rewriting my config in lua, and opened a pull request a couple of days ago.

Strokkur424:
Yeah that seems to be correct, thanks!

PR #1525, opened 2026-05-24
Configuring/Basics/{Dispatchers,Window-Rules}: document the refactored fullscreen state

kolayne:
Documents the refactored `fullscreen_state` dispatcher. This change is **NOT IMPLEMENTED YET**, this PR serves as the new **API proposal**.

The new API design is based on the following observations:
- Maximizedness is orthogonal to fullscreenness. Also, maximizedness only exists in the compositor, and not for clients: clients are always told that they are maximized. Its nature is more similar to floatingness. Thus, it now has its own dispatcher `hl.dsp.window.maximize` with a signature similar to `hl.dsp.window.float`
- The names `internal` and `client` are not intuitive (though, I am not sure that I did a good job selecting the new names either... I'm open for your suggestions)
- One desired functionality is to just unsync the internal and compositor fullscreen states. Thus, `hl.dsp.window.fullscreen_state` now has an optional parameter for that.

### DIFF
--- a/content/Configuring/Basics/Dispatchers.md
+++ b/content/Configuring/Basics/Dispatchers.md
@@ -100,8 +100,9 @@ A monitor. Can be:
 | `kill(window?)` | Kill the process owning the window with a `SIGKILL`. |
 | `signal({ signal, window? })` | Send a POSIX signal to the process owning the window. |
 | `float({ action?, window? })` | set a window's floating state. |
-| `fullscreen({ mode?, action?, window? })` | set a window's fullscreen state. `mode` can be "maximized" and "fullscreen". `action` can be `toggle`/`set`/`unset` |
-| `fullscreen_state({ internal, client, action?, window? })` | set a window's fullscreen state with more precision. `action` can be `toggle`/`set`/`unset`. See [Fullscreenstate](#fullscreenstate) |
+| `maximize({ mode?, window? })` | set a window's maximized state (like fullscreen but keeps gaps and bars). `mode` can be `on`, `off`, or `toggle` (default). |
+| `fullscreen({ mode?, window? })` | set a window's fullscreen state. `mode` can be `on`, `off`, or `toggle` (default). |
+| `fullscreen_state({ mode_compositor, mode_client, unsync?, window? })` | set a window's fullscreen state with more precision. each mode can be `keep`/`on`/`off`/`toggle`. See [Fullscreenstate](#fullscreenstate) |
 | `pseudo({ action?, window? })` | set a window's pseudotiling state. |
 | `move({ direction, group_aware?, window? })` | move a window in a direction. `group_aware = true` will put windows in/out of groups alongside the given direction. |
 | `move({ workspace, follow?, window? })` | move a window to a workspace |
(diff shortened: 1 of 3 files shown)


kolayne:
@vaxerski, please take a look and let me know if you agree with the new API!

content/Configuring/Basics/Dispatchers.md
```diff
@@ -268,24 +269,49 @@ Some props are expanded from their window rule parents which take multiple argum
...
 
-For example:
+As long as the two states are in sync (default, corresponds to the [`sync_fullscreen` dynamic effect](../Window-Rules#dynamic-effects)),
+the window recieves true information about its fullscreenness, and changes to either of the states are propagated to the other one.
+
+### `hl.dsp.window.fullscreen_state` dispatcher
+
+The `hl.dsp.window.fullscreen_state` dispatcher allows to decouple the two states.
+
+`mode_compositor` (formerly: `internal`) and `mode_client` (formerly: `client`) set the compositor state and client state, respectively.
```
wabarTM, line 290: `mode_compositor` can be just `compositor`, since we are already setting some toggle state, adding `mode_` is unnecessary. If it was like `monitor.cm` it would be understandable to have some kind of prefix

kolayne: I'm not sure. First, it makes sense for the `hl.dsp.window.fullscreen` and `hl.dsp.window.fullscreen_state` dispatchers to be consistent: `hl.dsp.window.fullscreen` has just one setting called `mode`, so `hl.dsp.window.fullscreen_state` has two settings, that is, `mode_compositor` and `mode_internal`. Second, I'm afraid that if we call it `compositor` then `hl.dsp.window.fullscreen_state({ ..., compositor = "off" })` will read exactly like `compositor = "off"`, implying something like "the compositor is off for this window" => ??

kolayne:
Hey, @vaxerski, this seems like a pretty large change to implement. I would appreciate it if you took a look at this proposal and confirmed if we're on the same page. It would be sad to discover that we have fundamental disagreements after I implement it...

content/Configuring/Basics/Window-Rules.md
```diff
@@ -60,13 +60,13 @@ The supported fields for the `match` table are:
 | tag | \[name\] | Windows with matching `tag`. |
 | xwayland | \[bool\] | Xwayland windows. |
 | float | \[bool\] | Floating windows. |
-| fullscreen | \[bool\] | Fullscreen windows. |
+| maximized | \[bool\] | Maximized windows. |
+| fullscreen | \[bool\] | Windows that are displayed fullscreen (corresponds to `mode_compositor` of [Fullscreenstate](../Dispatchers/#fullscreenstate)). |
+| clientwise_fullscreen | \[bool\] | Windows that are told they are fullscreen (corresponds to `mode_client` of [Fullscreenstate](../Dispatchers/#fullscreenstate)). |
```
vaxerski, line 65: client_wise?

content/Configuring/Basics/Dispatchers.md
```diff
@@ -100,8 +100,9 @@ A monitor. Can be:
 | `kill(window?)` | Kill the process owning the window with a `SIGKILL`. |
 | `signal({ signal, window? })` | Send a POSIX signal to the process owning the window. |
 | `float({ action?, window? })` | set a window's floating state. |
-| `fullscreen({ mode?, action?, window? })` | set a window's fullscreen state. `mode` can be "maximized" and "fullscreen". `action` can be `toggle`/`set`/`unset` |
-| `fullscreen_state({ internal, client, action?, window? })` | set a window's fullscreen state with more precision. `action` can be `toggle`/`set`/`unset`. See [Fullscreenstate](#fullscreenstate) |
+| `maximize({ mode?, window? })` | set a window's maximized state (like fullscreen but keeps gaps and bars). `mode` can be `on`, `off`, or `toggle` (default). |
+| `fullscreen({ mode?, window? })` | set a window's fullscreen state. `mode` can be `on`, `off`, or `toggle` (default). |
```
vaxerski, line 104: fullscreen action should be kept for compat reasons (doesnt have to be documented tho), just noting

content/Configuring/Basics/Dispatchers.md
```diff
@@ -100,8 +100,9 @@ A monitor. Can be:
...
+| `fullscreen_state({ mode_compositor, mode_client, unsync?, window? })` | set a window's fullscreen state with more precision. each mode can be `keep`/`on`/`off`/`toggle`. See [Fullscreenstate](#fullscreenstate) |
```
vaxerski, line 105: same here

content/Configuring/Basics/Window-Rules.md
```diff
@@ -60,13 +60,13 @@ The supported fields for the `match` table are:
 | tag | \[name\] | Windows with matching `tag`. |
 | xwayland | \[bool\] | Xwayland windows. |
 | float | \[bool\] | Floating windows. |
-| fullscreen | \[bool\] | Fullscreen windows. |
+| maximized | \[bool\] | Maximized windows. |
```
vaxerski, line 63: hm, atm fullscreen includes maximized. I dont think we can do this without a silent window rule break.

vaxerski (commented):
no opposition really as long as you keep compat and break as little as possible

content/Configuring/Basics/Window-Rules.md
kolayne: Hmm. We could have an extra rule to cover both. But it would not make sense to call it `fullscreen`.

How should we handle this?

content/Configuring/Basics/Dispatchers.md
kolayne: I don't understand. If the purpose of the dispatcher changes, and the `mode` argument now handles what `action` used to handle, how should we support `action`?

I was following the opposite logic: I changed it such that no old value of `mode` is a valid argument for the new parameter `mode`, making it so that an attempt to use the old dispatcher will necessarily trigger an error, forcing the user to update their config.

content/Configuring/Basics/Dispatchers.md
kolayne: Similar to `fullscreen`, the new `fullscreen_state` dispatcher is completely incompatible with the old one, and we even change the names of arguments. Why should we keep `action` around?

content/Configuring/Basics/Dispatchers.md
vaxerski: if action is specified use mode as it used to be used to not break config

content/Configuring/Basics/Window-Rules.md
vaxerski: no clue thats the thing

content/Configuring/Basics/Dispatchers.md
kolayne: 1. The `action` parameter is optional in the old dispatcher versions, so we its presence is not a reliable necessary condition for detecting that the config wants the old version of a dispatcher.
2. Why do you want to keep the new version compatible with the old one? I get that configs shall not break _silently_. But doesn't Hyprland keep it's major version at zero (i.e., `v0.*.*`) exactly to allow for breaking changes?

vaxerski: I mean if we can keep compatibility why not

PR #1534, opened 2026-05-29
Recommend uwsm users to enable hyprsunset as a service

### LINKED ISSUE #1141: hyprsunset: recommend uwsm users to enable as a service
I personally use uwsm to launch directly into hyprland, where I use both hypridle and hyprsunset. [Per the wiki](https://wiki.hypr.land/Hypr-Ecosystem/hypridle/#configuration), I have hypridle enabled as a user service (since it recommends that for uwsm users). 
Here is my hypridle.conf:
```
listener {
    timeout = 600
    on-timeout = hyprctl dispatch dpms off
    on-resume = hyprctl dispatch dpms on
}
```
If I start hyprsunset via hyprland.conf (`exec-once = hyprsunset`), upon resuming from idle, I get an error stating "could not connect to socket" if I use any hyprsunset temperature changing commands. I have to manually launch hyprsunset for the error to resolve.

I'm able to fix this issue by instead removing `exec-once = hyprsunset` from hyprland.conf and instead adding it as a service, [per the wiki](https://wiki.hypr.land/Hypr-Ecosystem/hyprsunset/). After rebooting, I'm able to idle and come back from idle with hyprsunset still working.
My problem is that the wiki should recommend uwsm users to launch hyprsunset as a user service like it does for hypridle instead of just saying it's an alternative to avoid this issue. Hope that makes sense.

Ansh2209:
The wiki mentioned the systemd service for [hyprsunset](https://wiki.hypr.land/Hypr-Ecosystem/hyprsunset/#usage) as an alternative, instead of recommendation for uwsm users unlike [hypridle's wiki](https://wiki.hypr.land/Hypr-Ecosystem/hypridle/#configuration).

Closes #1141

### DIFF
--- a/content/Hypr Ecosystem/hyprsunset.md
+++ b/content/Hypr Ecosystem/hyprsunset.md
@@ -77,7 +77,7 @@ profile {
 ## Usage
 
 To autostart hyprsunset, add: `hyprsunset` to your `hyprland.lua`'s autostart.
-Alternatively, use `systemctl --user enable --now hyprsunset.service` in order to use hyprsunset as a systemd user service.
+If Hyprland is started with [uwsm](../../Useful-Utilities/Systemd-start), you can use `systemctl --user enable --now hyprsunset.service`.
 
 Hyprsunset can also be controlled by supplying arguments to the command.  
 By specifying `hyprsunset --temperature 5000` you will override the current active config's temperature setting.  This however, will be overridden once a new profile is activated.


fufexan:
It's not necessary to use uwsm in order to use the systemd services. The change is technically correct, but not the only use-case for the systemd service.

vaxerski (approved):

vaxerski:
I dont think people care about sysd services for apps unless theyare  on uwsm

PR #1536, opened 2026-05-31
Configuring/Lua API: Added entire Lua API docs section

vinitlee:
Currently, Lua API documentation is spread across a few different locations and is embedded in more explanatory pages. It is sometimes difficult to simply see what types or table format a method expects or a class offers.

This change adds a "Lua API" section under Configuring which covers everything accessible under hl.* and all the HL.* types.


vaxerski:
how would this be more useful over LSP bindings? Most functions are self-explanatory.

vinitlee:
While the LSP bindings are helpful in development and the naming convention is clear, there are two key ways that the bindings leave a gap in usability.

First, information about loosely typed expected table shapes is lacking from the LSP stub. For example,

`hl.dsp.focus(...)`

Does not give information about accepted fields like `direction` or `monitor` and does not clarify what is expected in those fields. Whenever I encountered a spec lacking an explicit type when creating my configuration, I would need to seek the wiki and would need to try to figure out where the information was living. This format allows explanation of the expected fields.

Second is discoverability. Once you know about a method or class, it is relatively clear what it does, but knowing that it exists at all or that it uses a specific word or word order is not predictable. Having a tool that allows easy browsing of all available methods and classes and allows searching that may associate alternative names (scratchpad/special) bridges the gap. Personally, I have had a lot of "oh, I didn't realize I could use it to do that!" moments when using the proposed docs. This is especially important for supporting the creativity of hyprland users.

At the moment, these roles are attempted to be filled by supplementing the bindings with explanations on the wiki. The wiki is great for giving high level and structural explanations for topics, but the sections are broken up in a way that more lends itself to first time setup than reference.

vaxerski:
that's a fair point, but the formatting here is quite annoying, why split everything into files when you can just use md and a ToC is always on the right anyways?

PR #1561, opened 2026-06-13
Configuring/Advanced and Cool/Expanding-functionality.md: note that the int returned from window.active event is focus reason

erstarr:
Title.

Unless users look in the codebase, nowhere is this noted.


vaxerski:
isn't the int whether the window is active

erstarr:
> isn't the int whether the window is active

no it's `eFocusReason`.

e.g. https://github.com/hyprwm/Hyprland/blob/61d6ec78c6b5e0d80308867e8a3edd1b7e130cbb/src/desktop/state/FocusState.cpp#L211

vaxerski:
huh, right

PR #1573, opened 2026-06-24
Configuring/Advanced and Cool/Uncommon-tips-and-tricks: fix broken minimize windows

justkrysteq:
The current method to minimize windows with special workspaces does not work. I found a fix for it using tags.


fufexan:
I wonder how this plays with some bars that show workspaces. Would the named workspace show up persistently?

justkrysteq:
Well, most bars should ignore showing special workspaces, but I'm not really sure since I made one myself and had to set it up.

I love using this bind tho. I use Obsidian a lot and I like having it floating over my windows when I need to note something down, cuz it's always a bind away.

vaxerski (approved):

PR #1574, opened 2026-06-26
Configuring/Basics/Binds : reviewed the french layout instructions

GaetanQu:
Updated instructions for switching to AZERTY layout and provided a code example for binding keys, instead of redirecting to an obsolete guide.

### DIFF
--- a/content/Configuring/Basics/Binds.md
+++ b/content/Configuring/Basics/Binds.md
@@ -509,15 +509,22 @@ For instance, the [French AZERTY](https://en.wikipedia.org/wiki/AZERTY) layout u
 > [!NOTE]
 > To get the correct name for an `unmodified_key`, refer to [the section on uncommon syms](#uncommon-syms--binding-with-a-keycode)
 
+You can easily switch to AZERTY layout by editing this
 ```lua
--- On a French layout, instead of:
--- hl.bind(mainMod .. " + 1", hl.workspace(1))
-
--- Use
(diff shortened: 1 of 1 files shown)


content/Configuring/Basics/Binds.md
```diff
@@ -509,15 +509,22 @@ For instance, the [French AZERTY](https://en.wikipedia.org/wiki/AZERTY) layout u
...
+for i = 1, 10 do
+    local key = i % 10 -- 10 maps to key 0
+    hl.bind(mainMod .. " + " .. key,             hl.dsp.focus({ workspace = i}))
+    hl.bind(mainMod .. " + SHIFT + " .. key,     hl.dsp.window.move({ workspace = i }))
+end
+```
+to this
+```lua
+for i = 1, 10 do
+    local key = "code:" ..i + 9 -- 10 maps to key 0
```
vaxerski, line 523: literally wrong? also why would you do this?

GaetanQu: Well, it works well with me, since code:10 to code:20 are used to switch workspace on AZERTY layout. I just wanted to help configuring Hyprland easily, no need to be that rough

vaxerski: well the problem with azerty is that the un-shifted characters become not-numbers, which is awful.

But yeah I didn't notice the + 9, you should wrap that in a `()` to make it more clear. This is right then with that +9

PR #1610, opened 2026-07-31
Configuring/Basics/Window-Rules.md: Expand on the `pin` effect

kolayne:
Suggest to specify the `float` effect together with `pin`.


kolayne:
Btw, as far as I can see from the Hyprland code, at the time of mapping, the window can only be `float` in one of two cases:
- If requested explicitly with a window rule, or
- if requested by the XWayland manager.

So, all normal wayland windows are created non-floating, unless explicitly requested. I did not know that when I was writing a window rule for my picture-in-picture window, which always appears floating: I thought it is spawned as a floating window already, and I was pretty confused, when my rule with just a `pin = true` was not working.

Perhaps, it would be more intuitive for the `pin = true` effect to also implicitly enable the `float = true` effect. Maybe we can rename it to, e.g., `float_and_pin`. @vaxerski, what do you think?

vaxerski:
xwaylandmgr is named awfully, a wayland window can be floated by default too, eg if it locks min and max size to the same size. Thats probably pip. Check code

vaxerski (approved):

PR #1623, opened 2026-08-08
Configuring/Binds: warn that Lua keybind handlers must not block

### LINKED ISSUE #1622: Docs: warn that Lua keybind callbacks must not block
### Description

Lua keybind callbacks appear to run on the compositor main loop. Any blocking work inside a bind freezes the entire desktop until it returns.

I hit this with a custom Super+V "smart paste" that inspected the clipboard:

```lua
local h = io.popen("wl-paste --list-types 2>/dev/null")
local types = h:read("*a")
h:close()
```

When `wl-paste` hung (common on Wayland with empty/stale offers or XWayland sources), Hyprland stopped responding until the process was killed or the hang cleared. Stock Super+V that only injects keys (Shift+Insert / Ctrl+V) is fine.

This is expected event-loop behavior rather than a paste regression, but with Lua config it's an easy footgun. A short wiki note would help.

### Suggested docs note

On [Binds](https://wiki.hypr.land/Configuring/Basics/Binds/), next to the Lua function dispatcher example, something like:

> **Keybind handlers must not block.** Lua callbacks share the compositor event loop. Avoid `io.popen`, network I/O, clipboard tools (`wl-paste`, `xclip`), and other long-running work inside binds. Prefer `hl.dsp.exec_cmd(...)` for external commands, or a short-timeout wrapper if you must probe the system. Blocking the handler freezes input and the desktop.

### Optional feature request (Hyprland core, separate)

Documented or implemented non-blocking shell from Lua binds (fire-and-forget / callback), and/or a hard timeout around Lua bind execution so a hung external command cannot lock the compositor indefinitely. That would live in [hyprwm/Hyprland](https://github.com/hyprwm/Hyprland); this issue is for the wiki warning.

### System / repro

- Hyprland 0.56.2
- Wayland
- Repro: bind Super+V to `io.popen("wl-paste --list-types")` then paste after a clipboard offer that makes `wl-paste` hang

### Workarounds

1. Don't inspect the clipboard in the bind; inject keys only.
2. If you must probe: `timeout 0.1s wl-paste ...` so the hang is bounded (still not ideal on the main loop).
3. Use a separate chord for alternate paste behavior instead of branching on mime types at bind time.

I'll open a wiki PR shortly that adds the warning.

rlimberger:
## Summary

Adds a warning on the Binds page that Lua keybind callbacks must not block.

Lua bind handlers run on the compositor event loop. Calling blocking tools
from a bind (for example `io.popen("wl-paste --list-types")` for a "smart
paste") can hang forever when the clipboard offer stalls, which freezes the
whole desktop. Stock key injection is fine; the footgun is easy with Lua
config and was not documented.

## Changes

- After the Lua function dispatcher example on `Configuring/Basics/Binds`,
  add a WARNING callout: avoid `io.popen` / network / clipboard probes in
  binds; prefer `hl.dsp.exec_cmd(...)`; bound any necessary waits.

## Test plan

- [x] Matches existing wiki callout style (`> [!WARNING]`)
- [x] Placed next to the Lua function bind example where readers will see it
- [ ] Render with `hugo serve` if maintainers want a visual check

Fixes #1622

### DIFF
--- a/content/Configuring/Basics/Binds.md
+++ b/content/Configuring/Basics/Binds.md
@@ -33,6 +33,15 @@ hl.bind("SUPER + SHIFT + X", function()
 end)
 ```
 
+> [!WARNING]
+> **Keybind handlers must not block.** Lua callbacks run on the compositor
+> event loop. Avoid `io.popen`, network I/O, clipboard tools (`wl-paste`,
+> `xclip`), sleeps, and other long-running work inside bind functions.
+> Prefer `hl.dsp.exec_cmd(...)` for external commands so they run outside
+> the bind callback. If you must probe the system from Lua, bound the wait
+> (for example with `timeout`). A hung or slow call freezes input and the
+> entire desktop until it returns.
+
 ## Uncommon syms / binding with a keycode
 
 See the


vaxerski (approved):
there is a hard cap after which hyprland should kill lua, like 5s or so, but yes.

PR #1627, opened 2026-08-12
Describe built-in `graphical-session.target` management

LRitzdorf:
Goes with hyprwm/Hyprland#15776


LRitzdorf:
Pong @vaxerski now that the feature's merged in HL

PR #1634, opened 2026-08-18
Add French translation

kelian-4:
Hi,

This Pull Request proposes a French translation of the documentation.

Given its considerable size, reading the documentation in English can be quite challenging, even though I speak both languages. So, over the past few days, I took the time to create a French version of the documentation.

I used Claude to assist with the translation, but I personally reviewed and compared every file manually against the original English version to ensure the translation was accurate and consistent.

My goal is to make the documentation more accessible to French-speaking users, so they don't have to face the same issue I did.

I hope this translation can eventually be merged into the official documentation. If that happens, I will be happy to continue maintaining and updating the French version and the original documentation.

Thank you for taking the time to review this PR!


fufexan:
One thing I notice right away: the navbar wiki selector points to the English version. To work with the French page you'd need to edit `layouts/_partials/version.html` and `navbar-title.html`. The former should accept a language param and the latter should provide it via .Site.Language.Locale.

Not sure if we want translations, @vaxerski?

vaxerski:
we do not want translations this has been discussed in the past

kelian-4:
The fix is pushed, the version selector now respects the current language. @fufexan 🙂

And no worries about the comment on translations I didn't know it had already been discussed, no problem if it's not something you want to merge. Thanks for the code feedback anyway, now that I've read through the whole doc line by line I know Hyprland a bit better lol

> Une chose que je remarque tout de suite : le sélecteur wiki de la barre de navigation pointe vers la version anglaise. Pour travailler avec la page française, vous devrez modifier`layouts/_partials/version.html` et`navbar-title.html` . Le premier doit accepter un paramètre de langue et le second doit le fournir via .Site.Langue.Locale.
> 
> Je ne sais pas si nous voulons des traductions, @vaxerski?

PR #1640, opened 2026-08-29
Dang it, LaTeX

LRitzdorf:
Since we enabled LaTeX for math expressions, env vars (and their `$`) need to be codeblocked. That got skipped in one spot, on https://wiki.hypr.land/nix/configuring-hyprland-with-hjem/ :(

### DIFF
--- a/content/nix/configuring-hyprland-with-hjem.md
+++ b/content/nix/configuring-hyprland-with-hjem.md
@@ -3,8 +3,8 @@ weight: 30
 title: Configuring Hyprland with Hjem
 ---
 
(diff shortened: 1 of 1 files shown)


LRitzdorf:
Not really related, but FYI vax, the wiki still has old pages cached. Not that it's a massive deal, but having two wikis operational at once is a little weird

vaxerski (approved):

vaxerski:
> Not really related, but FYI vax, the wiki still has old pages cached. Not that it's a massive deal, but having two wikis operational at once is a little weird

I know, I'll clean the backend later

PR #1644, opened 2026-08-31
Update information on border_color window rule

RommelRivera:
For hyprwm/Hyprland#15581

### DIFF
--- a/content/configuring/core/rules/window-rules.md
+++ b/content/configuring/core/rules/window-rules.md
@@ -112,7 +112,7 @@ Dynamic effects are re-evaluated every time a property changes.
 | no_max_size | Removes max size limitations | bool | |
 | stay_focused | Forces focus on the window as long as it's visible | bool | |
 | animation | Forces an animation onto a window with an optional style (e.g., `"popin"`, `"popin 80%"`) | str | |
-| border_color | Sets the window's border to a color, gradient, or two gradients (active/inactive). See [border_color](#border_color-window-rule) below | gradient | |
+| border_color | Sets the window's border to a color or gradient. See [border_color](#border_color-window-rule) below | gradient | |
 | idle_inhibit | Sets an idle inhibit rule. Modes: `"none"`, `"always"`, `"focus"`, `"fullscreen"` | str | |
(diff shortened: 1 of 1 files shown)


wabarTM:
imo better create section below the table, describe it there and link in the description. PR adds 2nd parameter and another big example. also no fullstop at the end

wabarTM:
looks fine, but to not create more entities, better to use same formatting all over the place, like [here](https://wiki.hypr.land/configuring/core/animations/#general)

foobar we have:
- entry1
- entry2
- entry3

{{% details title="Examples" closed="true" %}}

\`\`\`lua
-- blablabla
code

-- s-udhrrbjf
code2
\`\`\`

{{% /details %}}

can you combine gradients with color in string variant? if so, add an example

p.s. 1 empty line around codeblocks

vaxerski:
@RommelRivera reminder about the above

RommelRivera:
Will do later today

content/configuring/core/rules/window-rules.md
```diff
@@ -172,6 +172,61 @@ The `group` effect takes a string with space-separated options:
...
+The `border_color` rule can receive the following:
+
+- A solid color or gradient (defaults to active border color)
+- Two solid colors or gradients
+- A solid color and a gradient
+- Or a lua table (only one color or gradient)
+
+{{% details title="Examples" closed="true" %}}
+
+```lua
```
vaxerski, line 180: I would not document those. We left them for compat, but they shouldn't be documented. Intended way is a single color

RommelRivera: Gotcha, so only lua table for colors then

vaxerski: single color or gradient table

RommelRivera: The single color still in lua format or string format as well?

vaxerski: single color lua format is just `"#ffeedd"`, wdym? it is string format

vaxerski (approved):

PR #1650, opened 2026-09-05
Configuring/Code snippets: fingerprint resistance for Tor Browser and Mullvad

ikv-ops:
Document how to keep Tor Browser and Mullvad Browser at their default window size.

Both browsers use one default window size so every user has the same fingerprint. Tiling resizes the window on open, and a non-default size makes the user stand out from the crowd. The rule floats the window so it keeps the default size.

### DIFF
--- a/content/configuring/code-snippets.md
+++ b/content/configuring/code-snippets.md
@@ -127,6 +127,27 @@ hl.on("window.open", function(w)
 end)
 ```
 
+### Open Tor Browser and Mullvad Browser at their default window size
+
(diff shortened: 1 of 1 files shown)


ikv-ops:
Ideally, this should become the config default.

vaxerski (approved):

ikv-ops:
@vaxerski thoughts on adding this to the default hyprland config? This will encourage more distros to ship this OOB, especially the ones who respect upstream choices.

vaxerski:
absolutely not

Issue #1333, opened 2026-01-01
Windowrules page examples are wrong.


addicust:
Windowrules page examples are wrong.

RVFET:
Yeah, same. My rules no longer work. I suspect it was the last update on arch two days ago, layerrules don't work either

La5u:
yeah its completely changed now. they removed windowrulev2, switched the syntax to Match, Rule instead of Rule, Match. The whole github wiki wrong 
https://wiki.hypr.land/Configuring/Window-Rules/

PR #1647, opened 2026-09-02
Post-rewrite fixups

LRitzdorf:
Stuff the wiki team (notably @Mausimux, @Aqa-Ib, @wabarTM, and me, plus @Dregu) noticed after #1616 got merged. Should just be small cosmetic tweaks.

PR #1653, opened 2026-09-08
configuring/core/rules/window-rules: fix move/size types, warn about rules with no effect

Mirilindo:
Two independent fixes to the Window Rules page.

### `move` and `size` are typed as `str`

`move`, `size`, `max_size` and `min_size` are all backed by the same
`CLuaConfigExpressionVec2` in `src/config/lua/bindings/LuaBindingsInternal.hpp`,
but only `max_size` and `min_size` were documented as `vec2`.
The examples already present in the `move` and `size` rows use table literals,
which contradicts their own type column.

The two tables also label that column differently -- "Type" under Static effects,
"Argument" under Dynamic effects -- which is probably how the two drifted apart.

Both forms are in fact accepted, so I noted that under "Expressions":

```lua
size = {800, 600}                          -- ok
size = "800 600"                           -- ok
size = {"monitor_w*0.5", "monitor_h*0.5"}  -- ok
```

### A rule with no effect is accepted silently

```lua
hl.window_rule({ name = "no-anim", match = { class = "kitty" } })  -- ok
hl.window_rule({ match = { class = "kitty" }, noanim = true })     -- error: unknown field 'noanim'
```

An unknown field is a hard error, but a rule that has a `match` and no effect
reloads cleanly and does nothing.
Since `name` is optional and free-form, a rule written as
`{ name = "no-anim", match = ... }` reads like it disables animations, passes
`hyprctl configerrors`, and has no effect.
I hit this twice in my own config before finding it.

The warning is placed at the top of the section, per the styling guide.
Related: hyprwm/Hyprland#15871 lists the same family of silently-accepted mistakes.

If you would rather fix the silence in the parser than document it, say so and I
will drop the second commit -- the first one stands on its own.

---

Verified on 0.56.2.
The type claim was re-checked against current `main` and is identical there.
The silent-no-op was only reproduced on 0.56.2, as I do not build from git --
please sanity-check that half.

On #1333: that issue predates the Lua rewrite, and its actual complaint
(`windowrulev2`, argument order) no longer applies to this page.
It looks stale and could probably be closed.


wabarTM:
> If you would rather fix the silence in the parser than document it, say so and I
will drop the second commit -- the first one stands on its own.

that is just dumb. if user didn't defined an effect, then he didnt defined an effect. there should be no warning about that

content/configuring/core/rules/window-rules.md
```diff
@@ -95,11 +114,14 @@ All position variables are monitor-local.
...
 - `cursor_x` and `cursor_y` for cursor position
 
+`move`, `size`, `max_size` and `min_size` all accept either a `{x, y}` table or a single space-separated string.
+
 Example expressions:
 
 ```lua
 move = {"window_w * 0.5", "(monitor_h / 2) + 17"}
 size = {"monitor_w * 0.5", "monitor_h * 0.5"}
+size = "800 600"
```
wabarTM, line 124: it is not an expression, you set sizes explicitly

content/configuring/core/rules/window-rules.md
```diff
@@ -73,13 +92,13 @@ This essentially means that it is always the `initialTitle` and `initialClass` w
 | group | Sets window group properties. See [group options](#group-window-rule-options) below | str |
 | maximize | Maximizes a window | bool |
 | monitor | Sets the monitor on which a window should open (e.g., `"1"`, `"DP-1"`). Can be suffixed with `" silent"` | str |
-| move | Moves a floating window to the given monitor-local coordinates (e.g., `{100, 200}`, `{"(cursor_x-(window_w*0.5))", "(cursor_y-(window_h*0.5))"}`) | str |
+| move | Moves a floating window to the given monitor-local coordinates (e.g., `{100, 200}`, `{"(cursor_x-(window_w*0.5))", "(cursor_y-(window_h*0.5))"}`) | vec2 |
```
wabarTM, line 95: as defined in naming, vec2 uses floats, but coords are ints, so it should be `table or string` instead

content/configuring/core/rules/window-rules.md
```diff
@@ -95,11 +114,14 @@ All position variables are monitor-local.
 - `window_w` and `window_h` for window size
 - `cursor_x` and `cursor_y` for cursor position
 
+`move`, `size`, `max_size` and `min_size` all accept either a `{x, y}` table or a single space-separated string.
```
wabarTM, line 117: if you are going to explain them here, remove examples from description, make this note a header instead and explain in details what you can do

wabarTM (changes requested):

wabarTM:
@Mirilindo waky-waky ai, go get the bag of flesh that hides behind you

Mirilindo:
Thanks for the review : the 4 points are addressed, and the PR is now one single commit.

**`vec2` (line 95)** -- I agree that `vec2` is not the right label, but I think floats are not the reason because they are accepted here. Both `size = {800.5, 600.5}` and `size = "800.5 600.5"` load without a config error on 0.56.2.

What actually differs is the parser. `move`, `size`, `max_size` and `min_size` use `CLuaConfigExpressionVec2` (`src/config/lua/bindings/LuaBindingsInternal.hpp`, lines 59-60 and 77-78), while a `vec2` option uses `CLuaConfigVec2`. I have measured this on 0.56.2 with `decoration.shadow.offset` as the `vec2` column :

| written | `vec2` | `move` / `size` / `max_size` / `min_size` |
| --- | --- | --- |
| `{800, 600}` | ok | ok |
| `{800.5, 600.5}` | ok | ok |
| `"800 600"` | ok | ok |
| `"800,600"` | ok | error |
| `{x = 800, y = 600}` | error | error |
| `"monitor_w*0.5 monitor_h*0.5"` | error | ok |

Ok for `table/string`. In #1647 `size` got this label but `max_size` and `min_size` were already labelled `vec2` on main and `move` is labelled `string`. So we need to relabel the other 3 ; otherwise the same inconsistency stays. Let me know if you'd prefer me to stick to `move`.

In #1647 `size = "200x500"` has been added, but Hyprland rejects it with `field 'size': expression vec2 requires two expressions separated by whitespace`. So `size = "200 500"` is OK.
This example is in the description cell of the `size` row.
It is moved out of the cell as you asked.
But it is not carried over into the new section because Hyprland rejects it.

**line 124** -- removed, you are right that it is not an expression.

**line 117** -- done in the way I understood it : the examples are out of the four description cells, which now link to a new `#### Position and size` section instead. It gives the 2 accepted forms, says where expressions may appear and lists what is rejected. That is the same shape #1644 just gave to `border_color`, so the page stays consistent. 
If I didn't understand it correctly, please let me know.

**second commit** -- dropped as offered when I opened the PR.

One thing that is not visible in the reviewed hunk: `max_size` and `min_size` take expressions too, through the same `calculateExpression` path as `move` and `size` in `WindowRuleApplicator.cpp`. Checked on 0.56.2 with a test window : `max_size = {"monitor_w * 0.1", "100"}` does clamp it. That is why the "Expressions" intro now names all four.

content/configuring/core/rules/window-rules.md
```diff
@@ -120,8 +145,8 @@ Dynamic effects are re-evaluated every time a property changes.
 | idle_inhibit | Sets an idle inhibit rule. Modes: `"none"`, `"always"`, `"focus"`, `"fullscreen"` | string | |
 | immediate | Forces the window to allow tearing | bool | |
 | keep_aspect_ratio | Forces aspect ratio when resizing with the mouse | bool | |
-| max_size | Sets the maximum size for floating windows (e.g., `{800, 600}`) | vec2 | |
-| min_size | Sets the minimum size for floating windows (e.g., `{200, 150}`) | vec2 | |
+| max_size | Sets the maximum size for floating windows. See [position and size](#position-and-size) | table/string | |
+| min_size | Sets the minimum size for floating windows. See [position and size](#position-and-size) | table/string | |
```
wabarTM:
```suggestion
| max_size | Sets the maximum size for floating windows. See [position and size](#position-and-size) | table or string | |
| min_size | Sets the minimum size for floating windows. See [position and size](#position-and-size) | table or string | |
```

content/configuring/core/rules/window-rules.md
```diff
@@ -73,20 +73,45 @@ This essentially means that it is always the `initialTitle` and `initialClass` w
 | group | Sets window group properties. See [group options](#group-window-rule-options) below | string |
 | maximize | Maximizes a window | bool |
 | monitor | Sets the monitor on which a window should open (e.g., `"1"`, `"DP-1"`). Can be suffixed with `" silent"` | string |
-| move | Moves a floating window to the given monitor-local coordinates (e.g., `{100, 200}`, `{"(cursor_x-(window_w*0.5))", "(cursor_y-(window_h*0.5))"}`) | string |
+| move | Moves a floating window to the given monitor-local coordinates. See [position and size](#position-and-size) below | table/string |
```
wabarTM:
```suggestion
| move | Moves a floating window to the given monitor-local coordinates. See [position and size](#position-and-size) below | table or string |
```

content/configuring/core/rules/window-rules.md
```diff
@@ -73,20 +73,45 @@ This essentially means that it is always the `initialTitle` and `initialClass` w
...
 | no_close_for | Makes the window uncloseable with `killactive` for a given number of ms on open | int |
 | no_initial_focus | Disables the initial focus to the window | bool |
 | pin | Pins the window (i.e. show it on all workspaces). _Note: pinning is ignored for non-floating windows. You most likely want to use this together with `float = true`_ | bool |
 | pseudo | Pseudotiles a window | bool |
 | scrolling_width | Set column width for window when starting on a workspace with the scrolling layout | float |
-| size | Resizes a floating window (e.g., `{800, 600}`, "200x500", `{"(monitor_w*0.5)", "(monitor_h*0.5)"}`) | table/string |
+| size | Resizes a floating window. See [position and size](#position-and-size) below | table/string |
```
wabarTM:
```suggestion
| size | Resizes a floating window. See [position and size](#position-and-size) below | table or string |
```

content/configuring/core/rules/window-rules.md
```diff
@@ -73,20 +73,45 @@ This essentially means that it is always the `initialTitle` and `initialClass` w
...
+max_size = "monitor_w*0.5 monitor_h*0.5"
+```
+
+The string form is split at the first whitespace, so write the expressions without spaces there.
+The table form has no such restriction.
+
+A single value (`size = 800`), a table with more or fewer than two elements, and the `{x = ..., y = ...}` form are all config errors, reported by `hyprctl configerrors`.
+Whitespace is also the only separator the string form accepts: `"800,600"` and `"800x600"` are both errors.
+A comma does work for [`vec2`](../../../naming-conventions#data-types) options, but these four are not `vec2`.
+
```
wabarTM, line 114:
```suggestion
<!-- TODO: Table arguments should be { x = int, y = int }, https://github.com/hyprwm/Hyprland/discussions/16339. When (i hope it is "when" and not "if") this is fixed, section below should be edited -->

Position and size arguments take a Lua table or a string.
Value can be a number or an [expression](#expressions).
The string argument is split at the first whitespace, so expressions should not contain spaces to be parsed correctly.

{{% details title="Examples" closed="true" %}}

```lua
-- Set size to x = 800, y = 600
size = { 800, 600 }

-- Same size, but via string
size = "800 600" 

-- Move window center to the cursor
move = { "cursor_x - (window_w * 0.5)", "cursor_y - (window_h * 0.5)" }

-- Set max_size as half of the monitor width and height
max_size = "monitor_w*0.5 monitor_h*0.5"
\```

{{% /details %}}
```

fucking github how do i do nested \`\`\` in a suggestion, without adding \\?! pls remove that one \ in a commit

content/configuring/core/rules/window-rules.md
wabarTM: Also, should that say "above"?

wabarTM (changes requested):
Apparently you cant just easily nest \`\`\` in suggestions

Mirilindo:
Thanks! Applied in ab687ab, with the `\` removed. I kept the `#### Position and size` heading (the four table rows link to it) and added one line listing the forms Hyprland rejects. And yes, "above", fixed.

wabarTM:
https://github.com/hyprwm/.github/blob/main/policies/AI_USAGE.md
image

hmmmmmm
@Mirilindo

Mirilindo:
Sorry for using AI. I hope my comments will still be useful to you.

PR #1616, opened 2026-08-06
Major wiki restructure, minor fixes, rewording

wabarTM:
Welp, that is a big-ass pr with a big-ass comment

Big wiki restructure/refactor because old wiki was adapted for hyprlang, and it was never actually addressed

Some changes:

 - Moved out types from variables to naming conventions
 - Duplicated "Hyprland is not meant to be a DE" warning in Installation page because people cant read
 - Merged example setups with preconfigured setups

 - Configuring: Start here -> Core
 - Configuring now consisnts of: Core, Lua snippets, Layputs, and Extra categories

 - Core: placed things in a "most important for a first time user" order, at least i tried to do so

 - Monitors: split big-ass page into more contained sub-pages

 - Variables:
   - Renamed to Config options
   - Added limits to what can be limited
   - Category and subcategory were renamed to Path because ["lua.can.do.that"]

 - Moved Devices into core, but additional properties are in need of clarification because they link to per-device binds page and per-device binds page links back to devices

 - Dispatchers's parameters where structured into: `(from/to) where? what (action)? how (to execute)?` order

 - Binds could be cleaner, but fully rewriting them is different pr
   - Removed workaround for binding only mod keysyms with release flag
   - Flags: split big-ass page into more contained sub-pages x2
   - Devices page refer to per-device binds page, and per-device binds page refer to devices page for `tags` property. I couldnt make it out so just left untouched

 - Rules: just rules, but split

 - Env vars: minor changes

 - Moved to core: Animations, autostart

 - Advanced configuration: consists of: lua events, lua utilities, hyprctl, notifications, permissions
   - events are evets
   - utils are helper functions. prep for a chore to actually documents them in the wiki instead of relying on luastub
   - hyprctl, notifs and perms are basically unchanged

 - Lua code snippets: basically collection of helpful functions and config snippets

 - Hypr ecosystem: this needs a separate pr, mainly just fixed links
 - Useful utils: same as ecosystem

 - Nix section changes by @Aqa-Ib:
   - Rewritten the main page of the section
   - Reorganized pages and content to be clearer
   - Added page for configuring with Hjem. Gave the user the choice to use hjem, home manager, or the upstream module for configuring hyprland
   - Refactored the hyprland on nixos installation page
   - Added info to the contributing page, manual building part
   - Some fixes and updates in general

And some minor updates and fixes for new structure in nvidia, ipc, crashes and bugs, faq, connect, contrib and debug

Issue #1600, opened 2026-07-19
Default to latest stable version instead of git


darkwater:
Most users will be on stable, not on the latest development version. Even with the small notice on the front page of the wiki, it's not clear and leads to plenty of confusion from people who come to the Discord to ask for help. (It has tripped me up a few times as well)

CaptaiNiveau:
Yeah same here, it'd be better to display a notice for an optional git version instead of the other way around.

wabarTM:
first page of wiki literally says that it is versioned
```
This wiki is versioned! By default, you are reading the wiki for the latest Hyprland git commit. Click on the version selector and select your version if you are running a tagged release (which you very likely are, you can check with hyprctl version).
```

CaptaiNiveau:
Yup, it could still show that and give you a link to an optional latest version that is only relevant to users on -git!

darkwater:
> first page of wiki literally says that it is versioned
> 
> ```
> This wiki is versioned! By default, you are reading the wiki for the latest Hyprland git commit. Click on the version selector and select your version if you are running a tagged release (which you very likely are, you can check with hyprctl version).
> ```

An explanation doesn't excuse bad design, but more specifically:
- You'll read this once when you're just getting started with Hyprland for the first time, at which point you've probably got other stuff to think about so you're immediately gonna forget that
- Wikis are often scanned, and that paragraph is just a regular bit of text in what looks like a "welcome!" opening text without its own header
- It's still stupid that git is the default, because it's not what most people are using

PR #1662, opened 2026-09-15
Use explicit versioned URLs, add quick selector in navbar

LRitzdorf:
The team working on the wiki rewrite (#1616) generally agreed that having a `/git/` segment in the URL would be smart, regardless of whether that remains the default version or not (#1600).

This PR adds a quick version selector/indicator in the navbar (@erstarr's idea) to take advantage of that, which should help contributors quickly swap between git and stable wikis. Depending on how others feel, this *might* pave the way for "latest stable" to become the default when visiting an unversioned link, but that's not the point of this PR.

To make this work, we'd need Vax to set up the following links/redirects on the backend:
- Move the current `/` submodule to `/git/`
- Symlink `/latest` to `/<actual-latest-release>`
- Redirect `/<anything-else>` → `/git/<anything-else>`, **this effectively sets the default version** and is subject to #1600
  - How hard is it to do an "else" redirect like this?

The backend repo is private, so unfortunately we can't PR that part ourselves.

Also, the version selector is removed from the sidebar (since it's linked in the navbar now), and the sidebar is hidden when on the version selector page.

Navbar screenshots:
- Git
  image
- 0.56.0
  image
- 0.56.0 marked as outdated
  image

Note that the latest stable version also needs to *not* have its `outdated` param set for this new header to work nicely.

cc @erstarr @Dregu @wabarTM


erstarr (approved):
Looks perfect, thank you.

vaxerski:
this is not possible with hugo because the prefix is expected by hugo at build time. I can't just "symlink latest" because then latest will forcefully redirect to e.g. v0.56.0 

I could build with "latest" into latest I guess...? But then what's the point of /git

LRitzdorf:
Ah, I was hoping a symlink would avoid doing a redirect on the user. Hm. Sounds like the backend is running with Hugo as the actual server process, rather than like an nginx or something (that we could configure more extensively)?

Edit: wait, "Hugo expects the prefix at build time," as in that prefix gets built into the links it writes? I'll check in a bit, but I thought all the cross-page links were done with relative paths specifically to avoid writing version-specific URLs for all of them

Edit 2 electric boogaloo: I guess the actually relevant question is, what's the backend repo structure, and how is it generated/served?

LRitzdorf:
Okay yeah, I think this might still be workable? What I've done locally is generate a second copy of the site into `public/0.56.0/`, and then create a symlink at `public/latest` pointing to `0.56.0`. That works; with the Hugo server running I can visit `localhost:1313/latest/` and get the contents of the 0.56.0 directory. That's all we'd need for `/latest` to be a valid URL and not a redirect.

The remaining problem is whether we can do "fallback" redirects, i.e. `wiki.hypr.land/thing` to `wiki.hypr.land/git/thing`. Those should be actual redirects to the browser, because the versioned URL (the second one there) is the link we'd want people to be copying and passing around

